Issue #2, opened 2015-03-12
Serializer doesn't special case abbreviations.


CyberSinh:
When I use the EventBuilder.AddObject method, with an object containing a dictionary<string, int>, I see on Exceptionless (Extended Data) that the key of the dictionary is changed (eg. "MP3" -> "m_p3", "FLAC" -> "F_l_a_c"). I use latest nuget package Exceptionless.Wpf.

Thanks

``` c#
ExceptionlessClient.Default.CreateLog("Source", "Message").AddObject(new LogInfo2()).Submit();
```

``` c#
public class LogInfo2
   {
      public LogInfo2()
      {
         files.Add("MP3", 281);
         files.Add("FLAC", 23);
      }

      private readonly Dictionary<string, int> files = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);

      public Dictionary<string, int> Files
      {
         get { return files; }
      }      
   }
```

niemyjski:
I'll take a look into this. It appears to be a bug in the ToLowerUnderscoredWords method:  

https://github.com/exceptionless/Exceptionless.Net/blob/fbdf657c968e8f5e819b1593304649e0ce8f2725/Source/Shared/Extensions/StringExtensions.cs#L18-L33

Is there any chance you could submit a unit test and patch for this?

CyberSinh:
Before trying to fix it, can you explain me what that method is supposed to do and in what context do you use it?

Thanks.

niemyjski:
It's supposed to break on Uppercase and and convert a worded string into lower case with a separator of _.

*Examples:*
```
BlakeNiemyjski -> blake_niemyjski
Blake   Niemyjski -> blake_niemyjski
Blake_ niemyjski -> blake_niemyjski
```

The issue is that it's not taking into account abbreviations like MP3 or IP Address.
```
MP3 -> mp3
IP Address -> ip_address
```

CyberSinh:
I fix this bug and sent you pull request (with fix and unit tests).

niemyjski:
Awesome, my only concern is that although the regex route is clean, it has a very good chance of being really slow compared to a for loop. This code will run to serialize every property of every event and needs to be lightning quick. Could you update the serializer tests to check the before and after implementation after 10k iterations of a complex event to see the impact?

niemyjski:
If there is an impact we'd need to convert this to a for loop, which might be better in the long run (thoughts?)

[...]

PR #3, opened 2015-03-14
Fix bug "naming issue with the serializer client side"

CyberSinh:


niemyjski:
Could you please create a branch for the submission client pull request and revert it from this branch (I have a task to test the naming pull request today). This makes it easier to accept individual pull requests. This normally wouldn't be an issue but this is a big change and I want to make sure it doesn't affect anything server side.

niemyjski:
If you can't that is fine, just preferred.

CyberSinh:
Done!

niemyjski:
I'm testing this server side and running into issues (can't even run the api). I'll need to investigate further.

niemyjski:
I think I found it.. 

We are getting the following exception:

```
A member with the name 'type' already exists on 'Nest.RootObjectMapping'. Use the JsonPropertyAttribute to specify another name.

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: Newtonsoft.Json.JsonSerializationException: A member with the name 'type' already exists on 'Nest.RootObjectMapping'. Use the JsonPropertyAttribute to specify another name.
```

If you look at the type:

```
    [JsonProperty("_type")]
    public ITypeFieldMapping TypeFieldMappingDescriptor { get; set; }

    [JsonProperty("type")]
    public override TypeNameMarker Type
    {
      get
      {
        return (TypeNameMarker) null;
      }
    }
```
We are splitting on `_` and we can't do that. I think we need to go character by character as before and not use the split. I also think that our setting needs to take the json property attributes into account as well.

niemyjski:
I think we also need to respect existing _'s we should handle the following cases:

`_type` -> `_type`
`_tYPe` -> `_tYPe`
`myCustom_type` -> `my_custom_type`
`myCustom  _type` -> `my_custom_type`

@ejsmith do you think if the string has a _ in it then we just use it? If so then the above would be

`_type` -> `_type`
`_tYPe` -> `_tYPe`
`myCustom_type` -> `myCustom_type`
`myCustom  _type` -> `myCustom  _type`

niemyjski:
@ejsmith do you think if the string has a _ in it then we just use it? If so then the above would be

CyberSinh:
So, what behavior should I implement in the code?

niemyjski:
Could you please update it to do the first method but allow for duplicate __ if they exist. I'd be more than happy to pair on this with you. We really appreciate your help.

We really like this change but it's a breaking change because it changes the names of event data in elasticsearch. So we need to be 100% sure we get this right when we change it because it's a massive amount of work to reindex data / ensure everything is working.

niemyjski:
I just updated the test cases.

CyberSinh:
Where is the updated test cases?

niemyjski:
_type -> _type
_tYPe -> _tYPe
myCustom_type -> my_custom_type
myCustom _type -> my_custom_type
node.data -> node.data
match_mapping_type -> match_mapping_type

CyberSinh:
Code changed as expected

niemyjski:
Thanks, I'll take a look soon

niemyjski:
I've yet to get a chance to look into this (I feel really bad), but it's on my task list.

CyberSinh:
Hi Blake, do you have any issue with this PR? Do you plan to merge it soon? Thanks.

niemyjski:
I have no issues with it, just need to do extensive testing with it. I hope to get to it soon. Been bogged down with support / javascript feature. I'll take a look at this soon (promise) in the mean time you could inject your own serializer that uses this code to take advantage of the naming. If you want I can meet up with you and show you how to do this.

niemyjski:
@CyberSinh Just and update. I'm working on this as well as working on getting rid of mongo. What this means is we are moving to only elastic search for storage which means one serialization setting. These will go hand in hand and will be tested. So the news is this will be merged really soon.

ejsmith:
@CyberSinh are you familiar with rebasing?  It would be good to get this pull request down to a single commit of only the things that have changed. If not, it's ok we can handle it.

CyberSinh:
@ejsmith sorry, I'm not familiar with rebasing.

ejsmith:
@niemyjski what is the plan with this pull request? It sounds like we were good with the change. Can you think of any issues?

niemyjski:
Here is a recap from: https://github.com/exceptionless/Exceptionless.Net/issues/2

@ejsmith @CyberSinh I just spent a few hours going over your change and we really like it and really really want it... We tried to implement it into our big server changes that are in coming. However, we may have to punt on this for a while. Here are the issues we are seeing after discussion and testing.

* We have millions of events stored and indexed in elastic search.
  * This change would cause anyone who has extended data indexed to have to do two searches to get at the same data (mp3 and m_p_3) as it's an unknown field so we wouldn't know how to handle it during reindexing.
  * We'd have to write a reindex script and run it to rename our fields like i_p_address so our models will deserialize, but the issue is, you can't use neither the old or new serializer while it's running (we'd have to figure this out further).
* We'll have to update all clients both JavaScript and .NET with the code changes (not a problem).
* We'll have to write a event migration script to rename json fields when we parse the event for all legacy events (easy to do just another thing to add to the task list).

niemyjski:
I'm closing this as we will be fixing this in the near future. We will be reindexing our server side data when we upgrade to elastic 5.1 as well as JSON.NET 9.0 adds a snake case data contract serializer out of the box.

Issue #12, opened 2015-04-13
.NET 4.0 build target


Phyxion:
Everything seems to use the Exceptionless.Portable package, which is a portable client package (as the name suggests). However, PCL is not supported on .NET 4.0 unless patched. Can you add another, non-portable, build target for .NET 4.0?

This seems related to https://github.com/exceptionless/Exceptionless.Net/issues/5

niemyjski:
We hadn't planned on on creating a non pcl core package, but will if required. Are you running into this often? How do you deploy your app? if it's msi based you can add a launch condition that requires this [patch](http://www.microsoft.com/en-us/download/details.aspx?id=3556) to be set. I know this isn't the best answer but we want to make it just work. We should never blow up...  If we did end up compiling as a full library what should we call it?

Phyxion:
Yes, we really need a non-PCL package as many users still use an unpatched .NET installation. Most installations go through inno setup but a major part is automatically updated and we cannot silently install the patch there. About the naming, it's not really 'Portable' anymore, but you can create another build target that targets just .NET 4.0 (Client Profile, preferably). For example, you can do so by adding another project (maybe also sln) and referencing the same files. You can code in one sln / project but if you add another file you would have to reference it in the other project as well. Maybe you can also add another property group in the same .csproj file, I know you can do this with .NET 4.0 and .NET 4.5 although I don't know if you can do this with PCL.

Getting the Portable to compile as .NET 4.0 Client Profile is very easy, it requires you to change the WebRequest.CreateHttp method (it does not exist in .NET 4.0) and change the using statements in two files (I do not recall which ones, it's about MemberTypes.Field). As references you need to add System, System.Net, System.Runtime.Serialization and System.Xml.

niemyjski:
I'll bring this up to the team for discussion. 

One thing we do in our production app is you can tell the msi installer that it requires specific things like minimum operating system / kb to be installed before your app will install. You could also target .NET 4.5 in your app and then it would be guaranteed to work. If you are running into this issue, I can guarantee you are hitting other issues on unpatched .net 4.0 machines. Since 4.5 is a sxs install, it actually modifies the behavior of your deployed .NET 4.0 app which is another thing to consider too.

Phyxion:
[...]

PR #13, opened 2015-04-15
Added .NET 4.0 target.

Phyxion:
--> https://github.com/exceptionless/Exceptionless.Net/issues/12

### DIFF
--- a/BuildAll.proj
+++ b/BuildAll.proj
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemGroup> 
+        <ProjectToBuild Include="Exceptionless.Net.sln">
+            <Properties>Configuration=Release;</Properties>
+        </ProjectToBuild>
+        <ProjectToBuild Include="Exceptionless.Net.sln">
+            <Properties>Configuration=Release;OutputPath=bin\Release40\;IntermediateOutputPath=obj\Release40\;UseNet40=true;</Properties>
+        </ProjectToBuild>
+    </ItemGroup>
+    <Target Name="Build">
+        <MSBuild Projects="@(ProjectToBuild)"/>
+    </Target>
+</Project>
--- a/Source/Shared/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
+++ b/Source/Shared/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
@@ -48,7 +48,7 @@ using System.Runtime.CompilerServices;
 using Exceptionless.Json.Utilities.LinqBridge;
 #else
 using System.Linq;
-
+using MemberTypes = Exceptionless.Json.Utilities.MemberTypes;
 #endif
 
 namespace Exceptionless.Json.Serialization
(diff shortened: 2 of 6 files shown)


Source/Shared/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
```diff
@@ -48,7 +48,7 @@
 using Exceptionless.Json.Utilities.LinqBridge;
 #else
 using System.Linq;
-
+using MemberTypes = Exceptionless.Json.Utilities.MemberTypes;
```
niemyjski: Any reason for this change?

Source/Shared/Newtonsoft.Json/Serialization/JsonTypeReflector.cs
```diff
@@ -39,6 +39,7 @@
 using System.Linq;
 #endif
 using System.Runtime.Serialization;
+using MemberTypes = Exceptionless.Json.Utilities.MemberTypes;
```
niemyjski: Any reason for this change?

Source/Shared/Submission/DefaultSubmissionClient.cs
```diff
@@ -94,7 +94,7 @@ public class DefaultSubmissionClient : ISubmissionClient {
         }
 
         private HttpWebRequest CreateHttpWebRequest(ExceptionlessConfiguration config, string endPoint) {
-            var request = WebRequest.CreateHttp(String.Concat(config.GetServiceEndPoint(), endPoint));
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(String.Concat(config.GetServiceEndPoint(), endPoint));
```
niemyjski: I don't think this works on portable. What issue were you seeing?

BuildAll.proj
```diff
@@ -0,0 +1,14 @@
...
+            <Properties>Configuration=Release;</Properties>
+        </ProjectToBuild>
+        <ProjectToBuild Include="Exceptionless.Net.sln">
+            <Properties>Configuration=Release;OutputPath=bin\Release40\;IntermediateOutputPath=obj\Release40\;UseNet40=true;</Properties>
+        </ProjectToBuild>
+    </ItemGroup>
+    <Target Name="Build">
+        <MSBuild Projects="@(ProjectToBuild)"/>
+    </Target>
+</Project>
```
niemyjski: We shouldn't need this file at all. We build everything by calling https://github.com/exceptionless/Exceptionless.Net/blob/master/Libraries/Build.ps1

Source/Shared/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
Phyxion: Yes, it won't compile on .NET 4.0 otherwise (same as below), it doesn't know which MemberTypes to use.

Source/Shared/Submission/DefaultSubmissionClient.cs
Phyxion: WebRequest.CreateHttp does not exist on .NET 4.0. This does and also works on portable:

https://msdn.microsoft.com/en-us/library/bw00b1dc%28v=vs.100%29.aspx
> Supported in: Portable Class Library

BuildAll.proj
Phyxion: Yes, I expected as much. You should 'migrate' this code to that file then.

niemyjski:
@Phyxion Can you please jump on gitter..

niemyjski:
committed this feature via https://github.com/exceptionless/Exceptionless.Net/commit/ee925cfa07662407b72b863b5a25dbde5478ff8d

Issue #37, opened 2015-07-10
Add support for reading configuration from environmental variables.


niemyjski:
When using git deploy you need to be able to read configuration from app settings or environmental variables.

PR #38, opened 2015-07-10
Pull request for  #37 Added the ability to read settings from environmental variables.

niemyjski:
I noticed there was a race condition that may be reintroduced here https://github.com/exceptionless/Exceptionless.Net/commit/da1f9f3ff00ea0fe8076d27d6a519811e2cf4ba0. However, I now think it's caused by the http modules calling startup multiple times and an exception happening.

@ejsmith :

1. Should we add checks to the modules to ensure init only happens once?
2. The environmental variable names are consistent with the app settings variables... Do you like this name?
3. Do you think we should drop enable ssl? To me it should be determined if you have https or http defined as part of the server url and we shouldn't be changing the url..

### DIFF
--- a/Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
+++ b/Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
@@ -176,7 +176,7 @@ namespace Exceptionless {
             
             string serverUrl = ConfigurationManager.AppSettings["Exceptionless:ServerUrl"];
             if (!String.IsNullOrEmpty(serverUrl))
-                config.ApiKey = serverUrl;
+                config.ServerUrl = serverUrl;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


niemyjski:
@ejsmith I added those checks (oops).. What do you think about points 1-3?

ejsmith:
1. Not sure what you mean.
2. Yeah, they are fine. We can't change the one that people are already using. So good to be consistent with that.
3. No, that is used in different ways.

niemyjski:
1. the Web Modules call startup repeatedly multiple times (I'm pretty sure) and that caused issues with the iss prestart stuff.
3. I show this is only used in one spot: https://github.com/exceptionless/Exceptionless.Net/blob/master/Source/Shared/Extensions/ExceptionlessConfigurationExtensions.cs#L141 and it could be inferred from the server url..

ejsmith:
I guess we used to be able to turn it off from the server side. So maybe it's not needed any more. We can't remove the config property though because people are using it.

niemyjski:
No but we could mark it as deprecated and remove the settings checks from the functionality I just added. It would allow us to remove it in the future :).

ejsmith:
Yeah, go for it.

niemyjski:
If you're happy merge the pull request and I'll do the rest :).

PR #41, opened 2015-07-17
add SubmitLog and CreateLog overloads using the LogLevel enum

xbelt:
I find this feature very useful. Choosing an option in an enum is easier than remembering the correct strings

### DIFF
--- a/Source/Shared/Extensions/ClientExtensions.cs
+++ b/Source/Shared/Extensions/ClientExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Exceptionless.Logging;
 using Exceptionless.Models;
 
 namespace Exceptionless {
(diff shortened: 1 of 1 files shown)


Source/Shared/Extensions/ClientExtensions.cs
```diff
@@ -63,6 +64,18 @@ public static class ClientExtensions {
...
 
         /// <summary>
+        /// Submits a log message event.
+        /// </summary>
+        /// <param name="client">The client instance.</param>
+        /// <param name="source">The log source.</param>
+        /// <param name="level">The log level.</param>
+        /// <param name="message">The log message.</param>
+        public static void SubmitLog(this ExceptionlessClient client, string source, string message, LogLevel level)
+        {
```
niemyjski: Can you please put the brackets on the same line (consistency) :).

Source/Shared/Extensions/ClientExtensions.cs
```diff
@@ -98,6 +111,24 @@ public static class ClientExtensions {
...
+        public static EventBuilder CreateLog(this ExceptionlessClient client, string source, string message, LogLevel level)
+        {
+            var builder = client.CreateLog(source, message);
+
+            var stringLogLevel = level.ToString();
+            if (!String.IsNullOrWhiteSpace(stringLogLevel))
+                builder.AddObject(stringLogLevel.Trim(), Event.KnownDataKeys.Level);
+
+            return builder;
+        }
```
niemyjski: I think this method should just return CreateLog(client, string, string, string) and call level.ToString() as the last argument. This way we don't have duplicated code that does the same thing :)

niemyjski:
Thanks for your pull request, I've made some comments :). Can you also please make sure you sign our CLA: http://www.clahub.com/agreements/exceptionless/Exceptionless

xbelt:
Thank you for your comments!
I needed to recreate the pull request because I was working on the master branch. I already signed the CLA.
So I'll close this pull request now

PR #42, opened 2015-07-18
create overloads for CreateLog and SubmitLog to accept the LogLevel enum

xbelt:
Same as previous pull request. Changed the syntax and removed unnecessary code. Had to put the code into a new branch so all changes were reflected in a single commit

### DIFF
--- a/Source/Shared/Extensions/ClientExtensions.cs
+++ b/Source/Shared/Extensions/ClientExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Exceptionless.Logging;
 using Exceptionless.Models;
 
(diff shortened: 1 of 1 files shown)


niemyjski:
@xbelt Looks good. 

@ejsmith I almost wonder if we should be putting a default log level on this overload but I guess it would make things inconsistent.

niemyjski:
@xbelt Thanks for your contribution!

xbelt:
@niemyjski I would refrain from putting a default log level on this overload. As you said it could make things inconsistent since there is already an overload of SubmitLog() which only requires a source and a message.

niemyjski:
I agree

Issue #8, opened 2015-03-29
Add expander showing actual Exception


azchrisf:
It would be great to have an expander button (either horizontal or vertical, vertical being proffered) that would expand the dialog and show the actual Exception that occurred in a read-only textbox.

This could include additional info such as Date/Time, machine name, etc.

This would be good for advanced and curious users.

niemyjski:
I think we should also have something like this. I think it would be good to show off all data that will be sent. Would you mind submitting a pull request for this? I think we should have a link button (click here to see...) and it shows a dialog with the content being sent.

azchrisf:
I'd love to help but I'm not a C# programmer, I do VB.NET. :-)
This seems like something that could be done in 15 minutes, I can give it a shot using a VB -> C# converter if you tell me what the method or property the UI uses to get the actual exception.

niemyjski:
I would just get our serializer (and all config.Resolver.GetJsonSerailizer() https://github.com/exceptionless/Exceptionless.Net/blob/master/Source/Shared/Dependency/DependencyResolverExtensions.cs#L61 ) and serialize it to a string and show that content as it's exactly what is sent. I'd add a link to the dialog that says something like Click here to show what's sent and then show the dialog

azchrisf:
I'm not sure how you implement that to get the data - I'm going to leave that department to you guys as you're more familiar with how your app works.

Here is the finished UI code

http://www.cfcsaz.com/personal/Dialog.zip

The only thing left is to populate the textbox with the Crash stack/info - AdditionalErrorInfo.Text
Let me know that snippet so I can throw it in there when your done.

niemyjski:
I've yet to get a chance to look into this, but it's on my task list.

niemyjski:
@xbelt  would you mind attaching a screenshot of the new behavior?

xbelt:
@niemyjski 
Here you go:
closed: 
![closed](http://puu.sh/j74F8/7a664e7864.png)
opened: 
![opened](http://puu.sh/j752u/664baafc2d.png)

niemyjski:
Thanks for the screenshots. Is there any way we can get these updated for wpf as well? @ejsmith thoughts?

ejsmith:
[...]

PR #44, opened 2015-07-18
fix issue #8

xbelt:
Fix for issue #8 
Later I will improve the code further but for now it works

### DIFF
--- a/Source/Platforms/Windows/Dialogs/CrashReportForm.cs
+++ b/Source/Platforms/Windows/Dialogs/CrashReportForm.cs
@@ -62,7 +62,11 @@ namespace Exceptionless.Dialogs {
 
         private void CrashReportForm_Load(object sender, EventArgs e) {
             _originalHeight = Height;
-            ErrorContentsBox.Text = Client.Configuration.Resolver.GetJsonSerializer().Serialize(Event);
+            try {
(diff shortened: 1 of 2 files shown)


Source/Platforms/Windows/Dialogs/CrashReportForm.cs
```diff
@@ -58,5 +59,24 @@ public sealed partial class CrashReportForm : Form {
             DialogResult = DialogResult.OK;
             Close();
         }
+
+        private void CrashReportForm_Load(object sender, EventArgs e) {
+            _originalHeight = Height;
+            ErrorContentsBox.Text = Client.Configuration.Resolver.GetJsonSerializer().Serialize(Event);
```
niemyjski: I wonder if we should put this in a try catch.. Just in case serialization fails.

xbelt:
Added the try catch and improved the form.
The screenshots will be available in the discussion about issue #8.

niemyjski:
Thanks! We've updated issue #8.

niemyjski:
I'm going to close this as it also needs to support wpf. Please feel free to open a pull request in the future with support for both (Ideally, we'd love if this opened a new dialog or layer ontop of this dialog with the content).

Issue #49, opened 2015-08-18
Feature request: OnSended event (WPF)


Stojo303:
When a user have submitted an issue from the submission dialog, it would be great to have an event so that the app can decide (for example) weather the applications execution should continue or if it should terminate.

Thanks in advance :)

niemyjski:
I don't think we should do an OnSended event but an OnQueued event that happens right after the event has been queued. This will allow you to exit or or process queue and then exit. By default the queue will be processed on wpf when the applications exists.

I think the handler should be something like this:

```csharp
OnQueued(IEnumerable<EventContexts> contexts)
```

This way you could see what was queued and take the proper action with the proper queue / client.

@ejsmith thoughts?

niemyjski:
I've created a pull request here: https://github.com/exceptionless/Exceptionless.Net/pull/51 Your feedback would be greatly appreciated.

niemyjski:
I'm not 100% sure if I'm happy with the way things are named. To me it gives the impression that the event has been sent to the server when it's just been queued (could even be an in memory queue). This event would allow people to close the application (WPF) if needed after an unhandled error.

niemyjski:
This has been fixed in pull request #51 and has been pushed: `3.1.1405`

niemyjski:
You'd just need to wire up to OnSubmitted Event handler and check the property to see if it's unhandled. If it is, close the application.

PR #51, opened 2015-08-21
#49 OnSubmitted Event Handler

niemyjski:
I'm not 100% sure if I'm happy with the way things are named. To me it gives the impression that the event has been sent to the server when it's just been queued (could even be an in memory queue). This event would allow people to close the application (WPF) if needed after an unhandled error.

### DIFF
--- a/Source/Shared/ExceptionlessClient.cs
+++ b/Source/Shared/ExceptionlessClient.cs
@@ -174,7 +174,7 @@ namespace Exceptionless {
                 _lastReferenceIdManager.Value.SetLast(ev.ReferenceId);
             }
 
-            OnSubmittedEvent(new EventSubmittingEventArgs(this, ev, pluginContextData));
+            OnSubmittedEvent(new EventSubmittedEventArgs(this, ev, pluginContextData));
         }
 
         /// <summary>Creates a new instance of <see cref="Event" />.</summary>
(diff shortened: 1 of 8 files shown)


Source/Shared/ExceptionlessClient.cs
```diff
@@ -225,6 +226,31 @@ public class ExceptionlessClient : IDisposable {
             }
         }
 
+        /// <summary>
+        /// Occurs when the event has been submitted.
+        /// </summary>
+        public event EventHandler<EventSubmittingEventArgs> SubmittedEvent;
```
ejsmith: This should probably be a different eventargs class since the submitting one has a cancel on it.

niemyjski: Okay, but does it make sense, because it hasn't been submitted (to our servers) but the submit method has been called so it's been submitted.

ejsmith:
I like this. I think it would be super handy to be able to do actions after the event has been submitted to the queue. Problem is that this event just means that it was put in the queue and if they wanted to do something like exit the app, they would need to make sure that they called the processqueue method.

niemyjski:
We'll that's not entirely true, because when wpf app closes it calls process queue.

Source/Shared/ExceptionlessClient.cs
ejsmith: Maybe we can add some additional options to this event handler to force the queue to be processed immediately.  Or maybe just add a ProcessQueue method right on the event args so it just makes it more discoverable. I'm not sure.

ejsmith:
That's true, but the use case we are talking about is them calling Application.Exit inside of this event handler.  Which means that code might not get called.

ejsmith:
I'm not even sure what this would be used for other than doing the exit on error. Any other use cases that you are aware of?

niemyjski:
I'm not aware of any other use case. I don't think we should add process queue to the submitted event args. It will happen when you have a desktop app automatically on process exit and can lead to abuse (calling it after every event submission).

niemyjski:
I just added a sample for this. I found that ProcessQueue may already be running in which case the app closes because ProcessQueue returns immediately. However, if the queue is not being processed then it processes the queue and shuts down.

ejsmith:
Hrmm... we might need to make it so that if process queue is currently running that processqueue will block until it's done.

niemyjski:
I really do t want to add locking to this code :(

niemyjski:
I'm going to merge in this change and if we get requests / feedback on this. Then we'll add it but we haven't gotten any complaints so far and the amount of issues adding locking to this code is unknown.

PR #54, opened 2015-08-31
Persistent Queue Fix

airwareic:
Fixes:
- Added configuration setting to disable deletion of queue items
- Changed deletion settings of queue items for non-persistentQueue settings to 7 days and 10 attempts

Bug Fixes:
- Changed dependency resolution for IsolatedStorage to get around registration exception thrown

- Various readability and simplification improvements, especially around upload attempt counting and file locking in the queue

### DIFF
--- a/Source/Extras/Exceptionless.Extras.csproj
+++ b/Source/Extras/Exceptionless.Extras.csproj
@@ -42,6 +42,7 @@
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Net" />
     <Reference Include="System.Net.Http, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.Net.Http.dll</HintPath>
--- a/Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
+++ b/Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
@@ -29,7 +29,7 @@ namespace Exceptionless {
         }
 
         public static void UseIsolatedStorage(this ExceptionlessConfiguration config) {
-            config.Resolver.Register<IObjectStorage, IsolatedStorageObjectStorage>();
+            config.Resolver.Register<IObjectStorage>(new IsolatedStorageObjectStorage(config.Resolver));
         }
 
         public static void UseFolderStorage(this ExceptionlessConfiguration config, string folder) {
@@ -94,6 +94,8 @@ namespace Exceptionless {
                 return;
 
             config.Enabled = section.Enabled;
+
+            config.PersistQueue = section.PersistQueue ?? true;
             
             if (IsValidApiKey(section.ApiKey))
                 config.ApiKey = section.ApiKey;
@@ -177,6 +179,10 @@ namespace Exceptionless {
             string serverUrl = ConfigurationManager.AppSettings["Exceptionless:ServerUrl"];
             if (!String.IsNullOrEmpty(serverUrl))
                 config.ServerUrl = serverUrl;
+
+            bool persistQueue;
+            if (Boolean.TryParse(ConfigurationManager.AppSettings["Exceptionless:PersistQueue"], out persistQueue))
+                config.PersistQueue = persistQueue;
         }
 
         /// <summary>
@@ -195,6 +201,10 @@ namespace Exceptionless {
             string serverUrl = GetEnvironmentalVariable("Exceptionless:ServerUrl");
             if (!String.IsNullOrEmpty(serverUrl))
                 config.ServerUrl = serverUrl;
+
+            bool persistQueue;
+            if (Boolean.TryParse(GetEnvironmentalVariable("Exceptionless:PersistQueue"), out persistQueue))
+                config.PersistQueue = persistQueue;
         }
 
         private static string GetEnvironmentalVariable(string name) {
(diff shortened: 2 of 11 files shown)


Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
```diff
@@ -29,7 +29,7 @@ public static class ExceptionlessExtraConfigurationExtensions {
         }
 
         public static void UseIsolatedStorage(this ExceptionlessConfiguration config) {
-            config.Resolver.Register<IObjectStorage, IsolatedStorageObjectStorage>();
+            config.Resolver.Register<IObjectStorage>(new IsolatedStorageObjectStorage(config.Resolver));
```
niemyjski: I don't like this change as it's been working. I'd like to have the stack trace as well as steps to reproduce this core issue before changing this.

Source/Extras/Exceptionless.Extras.csproj
```diff
@@ -42,6 +42,7 @@
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Net" />
```
niemyjski: Although this is more readable, all changes related to adding a reference to System.Net for readability need to be reverted. We initially went down this path but it doesn't work with pcl and some pcl configurations. There are two different versions of System.NET and there are runtime errors that can occur by using types against two different runtime versions. This is why we used int status codes.

Source/Shared/Queue/DefaultEventQueue.cs
```diff
@@ -74,35 +78,51 @@ public class DefaultEventQueue : IEventQueue {
...
-                        } else if (response.RequestEntityTooLarge) {
-                            if (batchSize > 1) {
-                                _log.Error(typeof(DefaultEventQueue), "Event submission discarded for being too large. The event will be retried with a smaller batch size.");
-                                batchSize = Math.Max(1, (int)Math.Round(batchSize / 1.5d, 0));
-                                deleteBatch = false;
-                            } else {
-                                _log.Error(typeof(DefaultEventQueue), "Event submission discarded for being too large. The event will not be submitted.");
+                        }
+                        else {
+                            switch ((HttpStatusCode)response.StatusCode)
```
niemyjski: I don't see any differences in this code other than the status code enum which we can't use due to the previous reasons :(. Can you revert this?

Source/Shared/Submission/SubmissionResponse.cs
```diff
@@ -3,11 +3,11 @@
 
 namespace Exceptionless.Submission {
     public class SubmissionResponse {
-        public SubmissionResponse(int statusCode, string message = null) {
+        public SubmissionResponse(int statusCode, string message) {
```
niemyjski: Any reason for not making this nullable?

Source/Shared/Extensions/FileStorageExtensions.cs
```diff
@@ -3,135 +3,310 @@
...
-        public static void DeleteBatch(this IObjectStorage storage, IList<Tuple<ObjectInfo, Event>> batch) {
+        public static void DeleteBatch(this IObjectStorage storage, IList<Tuple<ObjectInfo, Event>> batch)
+        {
             foreach (var item in batch)
                 storage.DeleteObject(item.Item1.Path);
         }
 
-        public static void ReleaseBatch(this IObjectStorage storage, IList<Tuple<ObjectInfo, Event>> batch) {
+        public static void ReleaseBatch(this IObjectStorage storage, IList<Tuple<ObjectInfo, Event>> batch)
+        {
```
niemyjski: Can you please revert all changes that are not related to the issue. I'm not trying to be nit picky and am greatful that users are doing pull requests but all our code shares style patterns and bracket changes makes things inconsistent and harder to look at the overall changes.

Source/Shared/Extensions/FileStorageExtensions.cs
```diff
@@ -3,135 +3,310 @@
...
         }
 
-        public static void ReleaseStaleLocks(this IObjectStorage storage, string queueName, TimeSpan? maxLockAge = null) {
+        public static void ReleaseStaleLocks(this IObjectStorage storage, string queueName, TimeSpan? maxLockAge = null)
+        {
             if (!maxLockAge.HasValue)
                 maxLockAge = TimeSpan.FromMinutes(60);
 
-            foreach (var file in storage.GetObjectList(queueName + "\\q\\*.x", 500).ToList().Where(f => f.Modified < DateTime.Now.Subtract(maxLockAge.Value)))
+            foreach (var file in storage.GetObjectList(GetAllLockedQueueItemsFilter(queueName)).Where(f => f.Modified < DateTime.Now.Subtract(maxLockAge.Value)))
```
niemyjski: We were getting at max 500 items here.

Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
ejsmith: Did you see an issue with this?

Source/Shared/Extensions/FileStorageExtensions.cs
ejsmith: Yes, we need the code style to match the rest of the project. It also makes it hard to see what has changed when there is a lot code style changes in the commit.

ejsmith:
Thank you for the contribution! I really appreciate you taking the time to do this. So the primary thing that you are trying to change is to make sure that we are not throwing away any events? Have you seen events being thrown away?  Let's discuss this more and thing figure out a plan of attack that we agree on.

Source/Extras/Extensions/ExceptionlessExtraConfigurationExtensions.cs
niemyjski: @airwareic was seeing a TinyIOC exception but I never saw the stack trace or have seen other reports. I'd love to see the stack trace to figure out what's going on.

PR #55, opened 2015-09-01
Offline Support and Persistent Queue

airwareic:
Added QueueMaxAttempts and QueueMaxAge configuration setting to allow custom persistence settings. This is intended for offline support.


ejsmith:
I like it!

PR #71, opened 2016-01-28
Deduplication

ejsmith:


Source/Shared/Models/Client/Event.cs
```diff
@@ -52,6 +52,39 @@ public class Event : IData {
         /// </summary>
         public string ReferenceId { get; set; }
 
+        /// <summary>
+        /// A unique id that identifies a usage session that this event belongs to.
+        /// </summary>
+        public string SessionId { get; set; }
```
ejsmith: FYI: SessionId is removed in the master branch.

Source/Shared/Plugins/Default/030_DuplicateCheckerPlugin.cs
```diff
@@ -1,40 +1,100 @@
...
 
-                // only keep the last 10 recent errors
-                Tuple<int, DateTime> temp;
-                while (_recentlyProcessedErrors.Count > 10)
-                    _recentlyProcessedErrors.TryDequeue(out temp);
+                // reset counter to be ready for the next wave
+                Count = 0;
 
-                current = current.Inner;
+                Client.SubmitEvent(Event, ContextData);
```
ejsmith: Thinking that we don't want to pass the ContextData through again because that will cause it to do work to gather all the data items for the event again.

Source/Shared/Plugins/Default/030_DuplicateCheckerPlugin.cs
```diff
@@ -1,40 +1,100 @@
...
+
+            public void IncrementCount() {
+                if (Count == 0) {
+                    // Handle the case after the reset.
+                    FirstOccurrence = DateTime.Now;
                 }
 
-                // add this exception to our list of recent errors that we have processed.
-                _recentlyProcessedErrors.Enqueue(Tuple.Create(hashCode, DateTime.Now));
+                Count++;
```
ejsmith: Probably want to do an Interlocked.Increment here.

Source/Shared/Models/Client/Event.cs
niemyjski: Yep, this needs to be removed.

niemyjski:
What's left for this pull request?

Source/Shared/Extensions/CollectionEqualityExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+                unchecked {
+                    var kvpHash = key.GetHashCode();
+                    kvpHash = (kvpHash * 397) ^ item.GetHashCode();
+                    keyValuePairHashes.Add(kvpHash);
+                }
+            }
+
+            keyValuePairHashes.Sort();
+            foreach (var kvpHash in keyValuePairHashes) {
+                hashCode = (hashCode * 397) ^ kvpHash;
```
srijken: Needs to be unchecked

niemyjski: What do you mean?

Source/Shared/Models/Client/Event.cs
```diff
@@ -43,6 +43,11 @@ public class Event : IData {
         public decimal? Value { get; set; }
 
         /// <summary>
+        /// The number of duplicated events
+        /// </summary>
+        public int? Count { get; set; }
```
niemyjski: Count isn't in the hashcode should it?

Source/Shared/Extensions/CollectionEqualityExtensions.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Exceptionless
+{
```
niemyjski: Can we fix the bracket formatting

Issue #73, opened 2016-01-29
Add plugins to ignore events that match the user agent filter


niemyjski:
We send down a client config setting called `@@UserAgentBotPatterns`. We should check this comma delimited list against the user agent..

PR #76, opened 2016-02-05
#73 Added plugin to ignore events that match a user agent bot filter

niemyjski:


niemyjski:
@ejsmith I added the ability to specify user agents as well client side.. The website talks about comma delimited list to match on.. But it should be wildcards / contains because the user agent could contains lots of data.

Please merge and close #73 if you like the changes I made

ejsmith:
Looks good!  Did you update the project settings UI?

PR #80, opened 2016-02-17
Fix geolocation for cultures that have a different separator for decimals

InlineAsm:
This fixes the geolocation for cultures like the Dutch culture that
separate decimals with a comma instead of a point.


niemyjski:
@InlineAsm did you run into any specific issues, the format we specify was putting a period instead of a space.

PR #82, opened 2016-02-23
Updates for custom stacking

niemyjski:


niemyjski:
@adamzolotarev thoughts?

niemyjski:
My only thoughts is that the model and known key should be changed from StackingInfo to ManualStackingInfo thoughts?

adamzolotarev:
Better than what I originally had. Thank you! Yeah, probably would better express the intent if it was called ManualStackingInfo... assuming it won't be used for anything else. But either way, look great to me!

PR #91, opened 2016-04-10
Set the event date to the largest date when deduplicating

srijken:


niemyjski:
Looks like there is conflicts now with my updates :(. There seems to be a race condition that is causing multiple issues that I've been running into. I added a lock and it seemed to help but something is causing the same event to have different hashcodes.

Source/Shared/Plugins/Default/1010_DuplicateCheckerPlugin.cs
```diff
@@ -102,6 +104,13 @@ private class MergedEvent {
                 _context.Event.Count = _count;
                 _context.Resolver.GetEventQueue().Enqueue(_context.Event);
             }
+
+            public void UpdateDate(DateTimeOffset date) {
+                lock (_lock) {
```
niemyjski: Would we still need this lock if we are locking on timer run and enqueue?

srijken: Well, the same could be said for Interlocked.Increment in IncrementCount()?

niemyjski: I don't think we need that lock there upon a second look. The outside is wrapped in it. I don't see anything wrong with interlocked.Increment (atomic)

PR #94, opened 2016-04-28
Added the ability to update a sessions activity without submitting an event.

niemyjski:


Source/Shared/Configuration/ExceptionlessConfiguration.cs
```diff
@@ -358,6 +378,9 @@ public class ExceptionlessConfiguration {
             if (String.IsNullOrEmpty(ServerUrl))
                 result.Messages.Add("ServerUrl is not set.");
 
+            if (String.IsNullOrEmpty(HeartbeatServerUrl))
+                result.Messages.Add("HeartbeatServerUrl is not set.");
```
niemyjski: Wonder if this will cause issues, maybe the ServerUrl setter should just set the heartbeat url by default? Otherwise configuring in code will require an extra setting.

PR #95, opened 2016-05-02
Feature/excludes

ejsmith:


Source/Shared/Extensions/StringExtensions.cs
```diff
@@ -29,13 +29,16 @@ public static class StringExtensions {
...
-            return patternsToMatch.Any(pattern => CheckForMatch(pattern, value, ignoreCase));
+            return patternsToMatch.Any(pattern => IsPatternMatch(value, pattern, ignoreCase));
         }
 
-        private static bool CheckForMatch(string pattern, string value, bool ignoreCase = true) {
+        public static bool IsPatternMatch(this string value, string pattern, bool ignoreCase = true) {
             if (pattern == null || value == null)
                 return false;
 
+            if (pattern.Equals("*"))
```
niemyjski: if the pattern is a star, wouldn't a null value match it?

ejsmith: I don't think it should.

ejsmith:
This looks good. I say we merge it and test it on the site.

PR #120, opened 2016-08-18
Added support for asp.net cores application lifetime

niemyjski:
http://www.khalidabuhakmeh.com/looking-at-asp-net-cores-iapplicationlifetime


ejsmith:
Looks good!

PR #123, opened 2016-09-17
Fixed a bug that was causing a format exception & proxy support

niemyjski:


niemyjski:
@ejsmith  can you please review this.

niemyjski:
We were getting a 

```
System.FormatException: Input string was not in a correct format.
at System.Text.StringBuilder.AppendFormatHelper(System.IFormatProvider provider, System.String format, System.ParamsArray args) at offset 0
at System.String.FormatHelper(System.IFormatProvider provider, System.String format, System.ParamsArray args) at offset 0
at System.String.Format(System.String format, System.Object[] args) at offset 32
at Exceptionless.Plugins.Default.DuplicateCheckerPlugin.Run(Exceptionless.Plugins.EventPluginContext context) at offset 55
at Exceptionless.Plugins.EventPluginManager.Run(Exceptionless.Plugins.EventPluginContext context) at offset 70
```
exception but I was unable to reproduce.

src/Exceptionless/Configuration/ExceptionlessConfiguration.cs
```diff
@@ -123,6 +119,16 @@ public class ExceptionlessConfiguration {
...
 
         /// <summary>
+        /// Used to identify the client that sent the events to the server.
+        /// </summary>
+        public string UserAgent { get; set; }
+
+        /// <summary>
+        /// Ability to set a custom proxy. By default, .NET will use any system or configuration defined proxy settings.
+        /// </summary>
+        public IWebProxy Proxy { get; set; }
```
niemyjski: I'm wondering if this should be apart of locked config. Once the client is created, this can't be changed.

src/Exceptionless/Submission/DefaultSubmissionClient.cs
```diff
@@ -15,10 +15,10 @@
 
 namespace Exceptionless.Submission {
     public class DefaultSubmissionClient : ISubmissionClient, IDisposable {
-        private readonly HttpClient _client;
+        private readonly Lazy<HttpClient> _client;
 
         public DefaultSubmissionClient(ExceptionlessConfiguration config) {
-            _client = CreateHttpClient(config.UserAgent);
+            _client = new Lazy<HttpClient>(() => CreateHttpClient(config));
         }
```
niemyjski: This will allow us to lazy create the client and allow people to create a custom implementation (this was running before the derived constructor was run).

ejsmith:
Looks good.

PR #131, opened 2016-11-29
FolderObjectStorage will only search in all directories if the search pattern starts with a wildcard

niemyjski:
https://bugzilla.xamarin.com/show_bug.cgi?id=42715


niemyjski:
I'm closing this due to this bug was merged into the mono project in December: https://github.com/mono/mono/pull/4042

Issue #77, opened 2016-02-05
Add support for X-Forwarded-For header


ejsmith:
Adding support for this will help with proxy servers and would enable us to properly identify individual users. Should at least make sure this is going to the server side to be analyzed there.

ejsmith:
https://en.wikipedia.org/wiki/X-Forwarded-For

niemyjski:
What would we need to set this to?

niemyjski:
We need to finish adding support for the other clients before we merge this into master

niemyjski:
We need to ensure we are removing the port from the IP address. I came across a use case where the port was submitted.

ejsmith:
Make sure that we don't lose any information though. So we shouldn't just be throwing the port away.

niemyjski:
We have to parse it correctly server side for geo ip lookup then, it's currently failing

PR #132, opened 2017-01-02
Client IP support for X-Forwarded-For added

barankaynak:
#77

### DIFF
--- a/src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
+++ b/src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
@@ -9,39 +9,29 @@ using Exceptionless.Extensions;
 using Exceptionless.Logging;
 using Exceptionless.Models.Data;
 
-namespace Exceptionless.ExtendedData
-{
-    internal static class RequestInfoCollector
-    {
+namespace Exceptionless.ExtendedData {
+    internal static class RequestInfoCollector {
         private const int MAX_DATA_ITEM_LENGTH = 1000;
 
-        public static RequestInfo Collect(HttpContextBase context, ExceptionlessConfiguration config)
-        {
+        public static RequestInfo Collect(HttpContextBase context, ExceptionlessConfiguration config) {
             if (context == null)
                 return null;
 
-            var info = new RequestInfo
-            {
+            var info = new RequestInfo {
                 HttpMethod = context.Request.HttpMethod,
                 UserAgent = context.Request.UserAgent,
                 Path = String.IsNullOrEmpty(context.Request.Path) ? "/" : context.Request.Path
             };
 
-            try
-            {
-                info.ClientIpAddress = GetIPAdres(context);
-            }
-            catch (ArgumentException ex)
-            {
+            try {
+                info.ClientIpAddress = context.Request.UserHostAddress;
+            } catch (ArgumentException ex) {
                 config.Resolver.GetLog().Error(ex, "An error occurred while setting the Client Ip Address.");
             }
 
-            try
-            {
+            try {
                 info.IsSecure = context.Request.IsSecureConnection;
-            }
-            catch (ArgumentException ex)
-            {
+            } catch (ArgumentException ex) {
                 config.Resolver.GetLog().Error(ex, "An error occurred while setting Is Secure Connection.");
             }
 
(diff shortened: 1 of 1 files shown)


src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
```diff
@@ -9,29 +9,39 @@
...
                 UserAgent = context.Request.UserAgent,
                 Path = String.IsNullOrEmpty(context.Request.Path) ? "/" : context.Request.Path
             };
 
-            try {
-                info.ClientIpAddress = context.Request.UserHostAddress;
-            } catch (ArgumentException ex) {
+            try
+            {
+                info.ClientIpAddress = GetIPAdres(context);
```
niemyjski: Address is spelled wrong, (should be `GetIpAddress(context)`)

niemyjski: Also this method should fallback to using (`context.Request.UserHostAddress`)

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
```diff
@@ -106,25 +136,48 @@ internal static class RequestInfoCollector {
...
+                {
                     if (!d.ContainsKey(key))
                         d.Add(key, "EXCEPTION: " + ex.Message);
                 }
             }
 
             return d;
         }
+
+        private static string GetIPAdres(HttpContextBase context)
```
niemyjski: Method name needs to be updated as shown above. Also this method should fallback to using (context.Request.UserHostAddress)

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
```diff
@@ -106,25 +136,48 @@ internal static class RequestInfoCollector {
...
+        {
+            string clientIp = context.Request.ServerVariables["HTTP_X_FORWARDED_FOR"];
+            if (!string.IsNullOrEmpty(clientIp))
+            {
+                string[] forwardedIps = clientIp.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                clientIp = forwardedIps[forwardedIps.Length - 1];
```
niemyjski: I'm not familiar with the spec, will the last ip address be the correct one? Any reason for not just joining all of them with `, `?

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
```diff
@@ -106,25 +136,48 @@ internal static class RequestInfoCollector {
...
+
+                string[] IPPort = clientIp.Split(new char[] { ':' }, StringSplitOptions.None);
```
niemyjski: What if the ip is Ipv6?

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
```diff
@@ -106,25 +136,48 @@ internal static class RequestInfoCollector {
...
+                clientIp = IPPort[0];
+            }
+            else
+            {
+                clientIp = context.Request.ServerVariables["REMOTE_ADDR"];
```
niemyjski: Is this the same as `context.Request.UserHostAddress`?

niemyjski:
Thanks for your PR, it's greatly appreciated. I added a review to the pr.

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
barankaynak: Yes, it is exactly same.
http://stackoverflow.com/questions/13994582/what-is-the-difference-between-request-userhostaddress-and-request-servervariabl

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
barankaynak: Ok, I will update it.

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
barankaynak: Ok, I will update it.

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
barankaynak: IPv6 support will be on the next pull request.

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
barankaynak: If user is behind a proxy, proxy adds its ip at the end of the string with a comma. It is ok if it is not a problem for storing it with multiple ips.

barankaynak:
I corrected the code by re-inspecting the topics you specified.

src/Platforms/Exceptionless.Web/RequestInfoCollector.cs
niemyjski: Thanks for your changes, they look good! Can we change the fallback to the context.request.userhostaddress since it's the same. By doing so we get compile time checking and don't have to worry about a string value / possible extra lookup?

barankaynak: Ok, I have updated.

barankaynak:
I think the latest commit meets the requirements. Can you please accept the pull request and update the nuget package? Thanks.

niemyjski:
Hey, I'll review it very soon and merge it into a branch and you'll get a prerelease package. We need to test this and also implement it for the other clients stacks before it goes into master.

PR #146, opened 2017-04-09
use the latest version of BenchmarkDotNet and disable shadow copy to …

adamsitnik:
Hi!

I have fixed the [bug](https://github.com/dotnet/BenchmarkDotNet/issues/395) that was blocking you from using BenchmarkDotNet.

It required following changes:

1. Use our CI NuGet feed to get the latest version
2. Disable shadow copy for the xunit

You can test it by running: `dotnet test -f net46 -c Release --filter "FullyQualifiedName=Exceptionless.Tests.Plugins.PluginTests.RunBenchmark"` from `Exceptionless.Net\test\Exceptionless.Tests` folder

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="appveyor-bdn" value="https://ci.appveyor.com/nuget/benchmarkdotnet" />    
+  </packageSources>
+</configuration>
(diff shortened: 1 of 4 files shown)


niemyjski:
Thanks for the pr, do you know if xunit.runner.json is still required?

adamsitnik:
I use `xunit.runner.json` because it worked with project.json files and now it works with the new .csprojs.

[xunit.github.io](https://xunit.github.io/) says that _"Configuring xUnit.net with JSON (all project types)"_ while _"Configuring xUnit.net with XML (desktop & PCL projects only)"_

You can also use   `"methodDisplay": "method"` to have short names in the test runner instead of the full names

niemyjski:
Thanks for the pr!!

Issue #147, opened 2017-04-12
00:41 Info  DuplicateCheckerPlugin - Event submission cancelled by plugin: refid= type= message=MYMESSAGE


MihaMarkic:
Who is this DuplicateCheckerPlugin guy and what does it want from me?

> 00:41 Info  DuplicateCheckerPlugin - Event submission cancelled by plugin: refid= type= message=MYMESSAGE

This is from UseInMemoryLogger and Exceptionless indeed doesn't send anything.

MihaMarkic:
Also note that my app is stuck with this error. Restart doesn't help. It also doesn't help that this error is silent (normally you wouldn't notice it unless logging is activated).

niemyjski:
@MihaMarkic We have a plugin (DuplicateCheckerPlugin) that sits in the pipeline and removes duplicate events. So for example event1 comes through we keep a hash of the entire event and mark it as sent. Lets say event2 comes in which is exactly the same as event 1, we cancel it and store it in a cache for 60 seconds. Then for the next 60 seconds any event that is exactly the same in every way is cancelled and updates event2's date and count property, so we know how many and when the last one has occurred. Then after the 60 second window we submit event2 with the exact count of how my occurred.

niemyjski:
So if you wait enough (60 seconds) you should see exact counts in the ui for that error.

niemyjski:
Are you running into an error where you believe it's discarding non unique events? Are you seeing the rolled up events in the ui?

MihaMarkic:
There is definitely something odd going on. Trying to pinpoint but it is like this:
Sometimes it won't send anything out, log would be empty (except for a single duplication warning). I produce like 20-30 logs in few seconds from more than one thread.
ExceptionlessClient.Default.SubmitLog is thread safe, right?

MihaMarkic:
It is a .net 4.6.1 console application.

MihaMarkic:
Also note that I issue ExceptionlessClient.Default.ProcessQueue(); before the end of the application.

niemyjski:
Yes, it should be thread safe. If you want, I would be more than happy to go on a remote meeting and take a look at this with you.

MihaMarkic:
@niemyjski I think I know what's going on. 

TL;DR; ProcessQueue() doesn't flushes always.

Rationale (I went looking into source code).
ProcessQueue would exit immediately when DefaultEventQueue.Process is running (it sets _processingQueue=true).
[...]

PR #148, opened 2017-04-13
Fixes ProcessQueue method so it blocks until queue has been processed

MihaMarkic:
ExceptionlessClient.Default.ProcessQueue() wouldn't block when Processing is under way.
The fix changes DefaultEventQueue.Process() by introducing an instance of Task while Process() is running (instead of **bool _processingQueue**).
If Process() method is running it'd return its Task when further calls are made. So the caller can await its completition.
Also removes unnecessary call to processingQueue within OnProcessQueue method.
This PR fixes issue #147.

### DIFF
--- a/src/Exceptionless/Queue/DefaultEventQueue.cs
+++ b/src/Exceptionless/Queue/DefaultEventQueue.cs
@@ -22,7 +22,7 @@ namespace Exceptionless.Queue {
         private DateTime? _suspendProcessingUntil;
         private DateTime? _discardQueuedItemsUntil;
 
-        public DefaultEventQueue(ExceptionlessConfiguration config, IExceptionlessLog log, ISubmissionClient client, IObjectStorage objectStorage, IJsonSerializer serializer) : this(config, log, client, objectStorage, serializer, null, null) { }
+        public DefaultEventQueue(ExceptionlessConfiguration config, IExceptionlessLog log, ISubmissionClient client, IObjectStorage objectStorage, IJsonSerializer serializer): this(config, log, client, objectStorage, serializer, null, null) {}
 
         public DefaultEventQueue(ExceptionlessConfiguration config, IExceptionlessLog log, ISubmissionClient client, IObjectStorage objectStorage, IJsonSerializer serializer, TimeSpan? processQueueInterval, TimeSpan? queueStartDelay) {
             _log = log;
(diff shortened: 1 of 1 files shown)


src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -75,43 +86,52 @@ public class DefaultEventQueue : IEventQueue {
                         var response = _client.PostEvents(events, _config, _serializer);
                         if (response.Success) {
                             _log.FormattedInfo(typeof(DefaultEventQueue), "Sent {0} events to \"{1}\".", batch.Count, _config.ServerUrl);
-                        } else if (response.ServiceUnavailable) {
+                        }
+                        else if (response.ServiceUnavailable) {
```
niemyjski: Could you please revert the formatting changes

src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -127,20 +147,26 @@ public class DefaultEventQueue : IEventQueue {
...
-            } catch (Exception ex) {
+            }
+            catch (Exception ex) {
                 _log.Error(typeof(DefaultEventQueue), ex, String.Concat("An error occurred while processing the queue: ", ex.Message));
                 SuspendProcessing();
-            } finally {
-                _processingQueue = false;
             }
+            finally {
+                tcs.SetResult(true);
```
niemyjski: should we be setting true or false based on if there was an error processing?

ejsmith:
Thanks Miha!

MihaMarkic:
@niemyjski Fixed formatting issues. Sorry I overlooked those. As per testing I did test with my code where I had issues. Admittedly I didn't unit test though as it seems difficult to. But the code looks more reliable than the previous one IMHO.

MihaMarkic:
@ejsmith You're welcome, I'm glad to contribute where I can.

src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -16,7 +16,8 @@ public class DefaultEventQueue : IEventQueue {
         private readonly IObjectStorage _storage;
         private readonly IJsonSerializer _serializer;
         private Timer _queueTimer;
-        private bool _processingQueue;
+        private Task _processingQueueTask;
+        private readonly object sync = new object();
```
ejsmith: This should be named _sync to match our naming conventions.

src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -45,20 +46,30 @@ public class DefaultEventQueue : IEventQueue {
...
             }
 
-            if (_processingQueue)
-                return;
+            TaskCompletionSource<bool> tcs;
+            lock (sync) {
+                if (_processingQueueTask != null) {
+                    return _processingQueueTask;
+                }
+                else {
```
ejsmith: Put the else on the same line to match our existing style } else {

ejsmith:
:shipit:

src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -45,20 +46,29 @@ public class DefaultEventQueue : IEventQueue {
         }
 
         public Task ProcessAsync() {
-            return Task.Factory.StartNew(Process);
+            return Task.Factory.StartNew(
```
niemyjski: I was thinking about this more, shouldn't this be the exact same code since it returns a task it should await on the task no? @MihaMarkic @ejsmith

MihaMarkic: I don't think so. Process returns a Task that has to be awaited.
In the original call, you'd invoke Process() method and ignore returned Task. So it will slip though.

Another way to solve it is to use something like
return Task.Run(async () => await Process());
but it feels like an overkill. Note that Task.Factory.StartNew ignores returned tasks (hence Task.Run)

MihaMarkic: Actually, Task.Run(() => Process()); should work as well. I'll modify it.

MihaMarkic:
Uf, I'm dense. Task.Run(Process); is the way to go.

src/Exceptionless/Queue/DefaultEventQueue.cs
```diff
@@ -45,20 +46,26 @@ public class DefaultEventQueue : IEventQueue {
         }
 
         public Task ProcessAsync() {
-            return Task.Factory.StartNew(Process);
+            return Task.Run(Process);
```
niemyjski: I do think this still needs to be start new so we don't block the current thread, it's kind of fire and forget.

ejsmith: That's what Task.Run does. It is a shortcut for Task.Factory.StartNew

niemyjski: Learned something new today, cool :)

niemyjski:
LGTM, thank you a ton for this pr, I'm going to merge it in and then make some more changes. You can grab a nightly build off myget. We'll need to do some additional testing before pushing the next release out.

src/Exceptionless/Queue/DefaultEventQueue.cs
MihaMarkic: AFAIK Task.Run creates a new Task, just like Task.Factory.StartNew does. The difference is that Task.Run will await for returned Task as well (if any), while Task.Factory.StartNew doesn't.

PR #156, opened 2017-06-14
Fix issue that the logging content cannot be recognized in Chinese environment.

edwardmeng:
The projects that we are using exceptionless client sdk with, have been developed and deployed in Simplified Chinese Windows environment. When there are some system errors occurred, the content in exceptionless.log file will be messy. So the encoding of logging writer should be changed to Encoding.UTF8 to compatible the variety cultures.


niemyjski:
Thanks!! I'll try and deploy a new build here soon with these changes. Is everything working okay currently?

Issue #158, opened 2017-07-28
.netstand2.0 System.AppDomain collision


majian159:
Error: CS0433
The type 'AppDomain' exists in both 'System.AppDomain, Version=2.0.11.0, Culture=neutral, PublicKeyToken=a741d64a8a4710e8' and 'System.Runtime.Extensions, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'

niemyjski:
We have a polyfill that needs to be removed for 2.0. Would you mind submitting a pr with a new framework target for 2.0 and remove the dependency (https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Exceptionless/Exceptionless.csproj#L52)

majian159:
This will bring other changes,I'm afraid to produce dirty code.
`            if (provider is ParameterInfo)
            {
                ParameterInfo p = (ParameterInfo)provider;
#if NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 && !NETSTANDARD2_0
                return ((attributeType != null) ? p.GetCustomAttributes(attributeType, inherit) : p.GetCustomAttributes(inherit)).ToArray();
#else
                return (attributeType != null) ? Attribute.GetCustomAttributes(p, attributeType, inherit) : Attribute.GetCustomAttributes(p, inherit);
#endif
            }`

niemyjski:
Yeah, I'm thinking maybe it might be best to rethink all those net standard attributes in the near future. It's find to produce some dirty code (for now), unless you have any ideas on how to clean that up. Maybe we could do NETSTANDARD && !NETSTANDARD2_0 (so we could replace all the if's with NETSTANDARD for the ones where it's all ors and just negate the ones that are not supported?

niemyjski:
Is the pr ready to be merged in? Is there anything I can do to help?

niemyjski:
@majian159 have you run into any issues with our nightly nuget package?

niemyjski:
I'm closing this as it's in master, just needs to be deployed.

zlzforever:
Could you please build a new package and upload to nuget? This bug cause my project build failed. Thanks a lot.

niemyjski:
[...]

PR #159, opened 2017-07-28
add support for .netstand2.0.

majian159:
#158 @niemyjski


majian159:
need to add AppVeyor support for .net core 2.0.

niemyjski:
It looks like they shipped all new 2.0 packages for asp.net core and netstandard 2.0. Would you mind including this? Also do you know if there is an open issue tracking netstandard 2.0 on appveyor (https://github.com/appveyor/ci/issues/1549)?

src/Exceptionless/Exceptionless.csproj
```diff
@@ -109,6 +109,27 @@
     <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0" />
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" Label="Build">
+    <DefineConstants>$(DefineConstants);NO_SQL_CLIENT;NETSTANDARD;NETSTANDARD1_5;NETSTANDARD2_0</DefineConstants>
```
niemyjski: Was there any specific reason for adding both 1.5 and 2.0 here? My guess is it cut down on the number of compiler directives needed greatly?

PR #161, opened 2017-08-09
Replace GetFiles with EnumerateFiles method to improve peformance in FolderObjectStorage

edwardmeng:
Replace GetFiles with EnumerateFiles method to improve peformance of FolderObjectStorage while retrieving object list. Especially in the high throughput scenarios, there will be more than thousands files, the performance of GetFiles will be very bad. So replace with EnumerateFiles can be very fast to retrieve file name with limit. Please refer to [https://stackoverflow.com/questions/5669617/what-is-the-difference-between-directory-enumeratefiles-vs-directory-getfiles](https://stackoverflow.com/questions/5669617/what-is-the-difference-between-directory-enumeratefiles-vs-directory-getfiles)

### DIFF
--- a/src/Exceptionless/Storage/FolderObjectStorage.cs
+++ b/src/Exceptionless/Storage/FolderObjectStorage.cs
@@ -121,7 +121,7 @@ namespace Exceptionless.Storage {
 
             var list = new List<ObjectInfo>();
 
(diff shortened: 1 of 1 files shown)


niemyjski:
Awesome find!! Thanks for the pr!

PR #162, opened 2017-09-07
Initial implementation of Microsoft.Extensions.Logging

moogle001:


niemyjski:
Thanks for the PR! I see that everyone else is doing the Product.Extensions.Logging which we should follow. Just kinda sounds like an extensions library for logging which I don't really like but will have to live with. Is there any chance you can update the asp.net core or console sample app as well in this pr so we can also see the usages of how we'd plug it in.

niemyjski:
@moogle001 is there any chance you'd be willing to update the sample console or asp.net core sample to use this logging library to test out?

src/Exceptionless.Extensions.Logging/ExceptionlessLoggerProvider.cs
```diff
@@ -33,6 +45,7 @@ public ILogger CreateLogger(string categoryName)
         public void Dispose()
         {
             _Client.Shutdown();
+            ((IDisposable)_Client).Dispose();
```
niemyjski: Not sure if we should be doing startup or shutdown here. Will log providers ever have scope and be disposed of in DI?

moogle001: Yeah, this is one of the questions I had for you. I had not realized the Exceptionless client was disposable until looking at the ASP code, and it appears the client's Dispose method does something different than the Shutdown method. Since the LoggerProvider creates its own client and does not expose it, the only way the client will ever be disposed is inside LoggerProvider Dispose method. On the otherhand, I'm not sure when/if that actually gets called in the Microsoft.Extensions.Logging life cycle.

The other question I had was, should new extension methods be added to Exceptionless.AspNetCore that work with the Microsoft.Extensions.Logging framework but also add the plugins and settings that are being configured when calling app.UseExceptionless(Configuration). Right now ExceptionlessAspNetCorePlugin and IgnoreUserAgentPlugin are not being added if loggerFactory.AddExceptionless((c) => c.ReadFromConfiguration(Configuration)) is used.

niemyjski: Typically what all our logging providers do is, it doesn't touch the client nor create one unless an api key is passed to the constructor of the provider (see nlog and the others). The only thing I'd do in dispose is call `client.ProcessQueue()` unless an api key was passed in (optional) and we created a client.

I'm almost thinking that we provide some overloads for logger factory, one that takes an api key, one that takes a client, one that takes nothing, and one that takes the config action... The one that takes a client will pass that through, the one that takes nothing will just use ExceptionlessClient.Default, and the other two will create a client instance only for logging.

https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs#L23. I'd rather keep it pretty simple like this, I think our other ones uses the default client or you pass in an api key.

moogle001: Alright, I adjusted the constructors and extensions. Let me know if it meets what you had in mind.

src/Exceptionless.Extensions.Logging/ExceptionlessLoggerProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExceptionlessLoggerProvider"/> class.
+        /// </summary>
+        /// <param name="configure">An <see cref="Action{ExceptionlessConfiguration}"/> which will be used to configure created loggers.</param>
+        public ExceptionlessLoggerProvider(Action<ExceptionlessConfiguration> configure) {
+            if (configure == null)
+                throw new ArgumentNullException(nameof(configure));
+
+            _Client = new ExceptionlessClient(configure);
+            _Client.Startup();
```
niemyjski: I'm thinking that if you pass in configure we shouldn't call startup as that subscribes to error handling and you may not want that. this would save the work of shutdown as well. I can make this change once I merge it in if you like.

niemyjski:
LGTM, Thanks for your commit and help! I'll get this merged in very soon as apart of our net standard 2.0 release :)

src/Exceptionless.Extensions.Logging/ExceptionlessLoggerProvider.cs
moogle001: If Startup isn't necessary, then by all means remove it!

moogle001:
Any idea when this will be published? Looking forward to using it.

niemyjski:
Soon, it's available right now on our myget feed. There is a lot to test in this point release (serialization changes, net standard 2.0 support, and many other changes. If you find any issues please let us know we could use some extra hands.

PR #166, opened 2017-10-26
Abstract serialization/deserialization of events. Fix #163

### LINKED ISSUE #163: Abstract serialization/deserialization of events and settings.
In the high load environment, the json serialization/deserialization became the bottle neck of performance, There will be too many temporary files kept. Now there is more than 10G, and increased day by day.
So could you please make the serialize/deserialize abstraction, that we can replace the json.net with MessagePack  or ProtoBuf to fix this problem.

edwardmeng:


src/Exceptionless/Serializer/IEventSerializer.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.IO;
+using Exceptionless.Models;
+
+namespace Exceptionless.Serializer {
+    public interface IEventSerializer{
```
ejsmith: I'd like to keep this a little more generalized. I think we should do something like this:

```c#
public interface IStorageSerializer {
   void Serialize<T>(T data, Stream output);
   T Deserialize<T>(Stream input);
}
```

edwardmeng: Currently the interface only used to serialize/deserialize events, whether it should be used for settings  in `FolderObjectStorage` if it is generalized as `IStorageSerializer`?

ejsmith:
Looks good! Will be great to be using the same serializer for the settings as well and also give us the ability store more things using the same mechanism. Thank you @edwardmeng!

niemyjski:
Thanks for the PR!

PR #167, opened 2017-11-02
Add event time and level for the file logger.

edwardmeng:
The original file logger cannot record event time and level. It's hard to know when the error occurs, and which one is warn or error.

### DIFF
--- a/src/Exceptionless/Logging/FileExceptionlessLog.cs
+++ b/src/Exceptionless/Logging/FileExceptionlessLog.cs
@@ -135,7 +135,7 @@ namespace Exceptionless.Logging {
                                 LogEntry entry;
                                 while (_buffer.TryDequeue(out entry)) {
                                     if (entry != null && entry.LogLevel >= MinimumLogLevel)
-                                        writer.Value.WriteLine(entry.Message);
+                                        writer.Value.WriteLine($"{FormatLongDate(entry.Timestamp)} {entry.LogLevel.ToString().PadRight(5)} {entry.Message}");
                                 }
                             }
                         } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -135,7 +135,7 @@ public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
                                 LogEntry entry;
                                 while (_buffer.TryDequeue(out entry)) {
                                     if (entry != null && entry.LogLevel >= MinimumLogLevel)
-                                        writer.Value.WriteLine(entry.Message);
+                                        writer.Value.WriteLine($"{FormatLongDate(entry.Timestamp)} {entry.LogLevel.ToString().PadRight(5)} {entry.Message}");
```
niemyjski: Could we use existing date time formats and place a formatter here?

src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -249,8 +279,10 @@ private class LogEntry {
             public LogEntry(LogLevel level, string message) {
                 LogLevel = level;
                 Message = message;
+                Timestamp = DateTime.Now;
             }
 
+            public DateTime Timestamp { get; set; }
```
niemyjski: I think most of our dates follow DateUtc, but timestamp is more log specific.

src/Exceptionless/Logging/FileExceptionlessLog.cs
edwardmeng: All the intrinsic date time formats are not kindly for reading, so the `FormatLongDate` method transported from NLog implementation, please refer to https://github.com/NLog/NLog/blob/master/src/NLog/LayoutRenderers/LongDateLayoutRenderer.cs

src/Exceptionless/Logging/FileExceptionlessLog.cs
edwardmeng: This date will be rendered to local log files, so I think it should be local date. If it's UTC date, the readers(mankind) will be confused since the timezone problem. And I have noticed the `InMemoryExceptionlessLog` uses the local date too, and the NLog uses the local date too.

niemyjski:
LGTM, if we ever want to change these values to utc, it shouldn't break anything as it's internal :)

PR #168, opened 2017-11-07
Add project to serialize/deserialize events and settings with MessagePack

edwardmeng:
Improve the serialization/deserialization performance of the object storage with MessagePack, please refer to #163.


edwardmeng:
@niemyjski I have finished unit tests, samples and bug fix, pls review the changes.

src/Exceptionless/Dependency/DependencyResolver.cs
```diff
@@ -24,6 +24,8 @@ public class DependencyResolver {
         }
 
         public static void RegisterDefaultServices(IDependencyResolver resolver) {
+            resolver.Register(resolver);
```
niemyjski: Was there a specific issue where the resolver was required but not registered?

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
```diff
@@ -1,10 +1,46 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace Exceptionless {
     internal static class CollectionEqualityExtensions {
+        private static bool ElementEquals(object source, object other) {
```
niemyjski: I'm a bit hesitant about these changes, was there an issue you ran into where these were required? I know that @srijken ran into a lot of performance issues with equality checks as they add a lot of overhead especially when doing duplicate checking.  @srijken do you have anything to add?

src/Exceptionless/Models/Client/Event.cs
```diff
@@ -80,7 +80,10 @@ public class Event : IData {
                 hashCode = (hashCode * 397) ^ (Tags == null ? 0 : Tags.GetCollectionHashCode());
                 hashCode = (hashCode * 397) ^ (Message == null ? 0 : Message.GetHashCode());
                 hashCode = (hashCode * 397) ^ (Geo == null ? 0 : Geo.GetHashCode());
+                hashCode = (hashCode * 397) ^ (ReferenceId == null ? 0 : ReferenceId.GetHashCode());
+                hashCode = (hashCode * 397) ^ Date.GetHashCode();
                 hashCode = (hashCode * 397) ^ Value.GetHashCode();
+                hashCode = (hashCode * 397) ^ Count.GetHashCode();
```
niemyjski: @srijken can confirm, but we didn't include count and date in the equalities because it breaks duplication checking. I'd feel more comfortable if these changes were discussed in an issue and put in a different pr. I'm thinking these were added because were checking equality in tests? But if we are round tripping they should be equal / we could check specific properties to ensure they are the same (better to do that than to pay this cost for every execution)

src/Exceptionless/Serializer/DefaultJsonSerializer.cs
```diff
@@ -26,18 +26,15 @@ public class DefaultJsonSerializer : IJsonSerializer, IStorageSerializer {
         }
 
         public virtual void Serialize<T>(T data, Stream outputStream) {
-            using (var writer = new StreamWriter(outputStream)) {
-                writer.Write(Serialize(data));
-            }
+            var buffer = System.Text.Encoding.UTF8.GetBytes(Serialize(data));
```
niemyjski: What issues did you run into where we had to convert this to bytes? This has been working for a long time

src/Platforms/Exceptionless.MessagePack.Signed/Exceptionless.MessagePack.Signed.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\..\build\common.props" />
   <PropertyGroup Label="Package">
-    <PackageId>Exceptionless.Log4net.Signed</PackageId>
-    <AssemblyName>Exceptionless.Log4net.Signed</AssemblyName>
+    <PackageId>Exceptionless.MessagePack.Signed</PackageId>
```
niemyjski: Let's completely remove the signed packages unless you are using them. We'll have to deprecate them in the very very near future and it stinks having to list them and delist them from nuget.

test/Exceptionless.MessagePack.Tests/Exceptionless.MessagePack.Tests.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net462</TargetFramework>
```
niemyjski: Can these run on .net standard or .net core?, if so we'd rather have all tests run on .net core as our long term goal is to focus on xplat and running them xplat

test/Exceptionless.MessagePack.Tests/Exceptionless.MessagePack.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <AssemblyName>Exceptionless.Tests</AssemblyName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="xunit" Version="2.3.1" />
```
niemyjski: The way the new project system works is dependencies of referenced projects are brought in automatically (transitive) so we don't need to define them more than once :)

test/Exceptionless.Tests/Serializer/SerializerTests.cs
```diff
@@ -11,7 +12,7 @@
 
 namespace Exceptionless.Tests.Serializer {
     public class SerializerTests {
-        protected virtual IJsonSerializer GetSerializer() {
+        protected virtual DefaultJsonSerializer GetSerializer() {
```
niemyjski: Perhaps we should rename this to JsonSerializer Tests and have this return the interface and have a second one for  Storage Serializer Tests that way the test base only returns the interface and the tests can run off that (easier to add tests for new implementations). We could also just case the serializer to our ISettingsSerializer and return if it doesn't match as well? What are your thoughts?

src/Exceptionless/Extensions/ReflectionExtensions.cs
```diff
@@ -16,7 +16,7 @@ internal static class ReflectionExtensions {
         public static DateTime? GetCreationTime(this Assembly assembly) {
 #if NET45 || NETSTANDARD1_5 || NETSTANDARD2_0
             try {
-                return File.GetCreationTime(assembly.Location);
+                return File.GetCreationTime(assembly.Location).ToUniversalTime();
```
niemyjski: It looks like we are only using this to set module info created and modified time. We've always stored it in local time, but I guess it probably should be in UTC because we are not storing the offset.. We should be calling GetCreationTimeUtc GetLastWriteTimeUtc instead.

src/Exceptionless/Models/Client/Event.cs
srijken: Indeed, especially Date shouldn't be included there. The Date of each event is different, and this will definitely break deduplication

src/Exceptionless/Serializer/DefaultJsonSerializer.cs
srijken: I think this also doubles memory allocation?

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
srijken: I can't say if this will be better or worse. We need to do perf benchmarks on this before and after these changes. Why were they needed in the first place, the PR is about WebPack?

srijken: I can't say if this will be better or worse. We need to do perf benchmarks on this before and after these changes. Why were they needed in the first place, the PR is about WebPack?

src/Exceptionless/Models/Client/Event.cs
srijken: Indeed. The Date of each event is different, and checking Date for comparison will definitely break deduplication.

Count is being determined by the deduplication plugin. Suppose events are being generated fast enough to trigger deduplication, then the aggregated event will have a count, and the new events coming in won't (they increase the aggregated event's count).

src/Exceptionless/Dependency/DependencyResolver.cs
edwardmeng: The `MessagePackStorageSerializer` has only constructor with parameter `IDependencyResolver` as following:
```cs
public MessagePackStorageSerializer(IDependencyResolver resolver) {
      _formatterResolver = new ExceptionlessFormatterResolver(resolver);
}
```
At the same time, the storage serializer can only registered by parsing configuration section with automatically constructor dependency, as following
```cs
if (section.StorageSerializer != null) {
      if (!typeof(IStorageSerializer).GetTypeInfo().IsAssignableFrom(section.StorageSerializer)) {
           config.Resolver.GetLog().Error(typeof(ExceptionlessConfigurationExtensions), $"The storage serializer {section.StorageSerializer} does not implemented interface {typeof(IStorageSerializer)}.");
      }
      else {
            config.Resolver.Register(typeof(IStorageSerializer), section.StorageSerializer);
      }
}
```
So if the `IDependencyResolver` is not registered, the `MessagePackStorageSerializer` cannot be constructed correctly in object storage. There will be exception thrown.

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
edwardmeng: @niemyjski In my understanding of the implementation of the `DuplicateCheckerPlugin`, only the `GetHashCode` method has been used, but not `Equals` method. Although the performance of `GetHashCode`  is better than `Equals`, but its not accurate for comparison(I have added unit test method refer to https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/8b4cf1e6109710ebec438b470ceca98c9e069374).
The `ElementEquals` method is used to check events that the value of data dictionary is `List<T>` or `Dictionary<K,V>`. I have added unit test method for this scenario, please refer to https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/c88b7bedbcdadd0b48822aecb25e97f134c66b80
@srijken I think the performance improvement should be based on the features implemented without issues. The previous unit test methods will be all failed, which I wanna create issues for them separately.

src/Exceptionless/Extensions/ReflectionExtensions.cs
edwardmeng: It's my fault, and I have committed changes in https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/f7efcb2efaf63c4bd9a2add79e888cd9f237da83

src/Exceptionless/Models/Client/Event.cs
edwardmeng: The `Event` is public and every properties can be visible for application developers, so they will assume that all the properties can be assigned and the `Equals` and `GetHashCode` should include all the property values too.
But in the library design purpose, the `Count` property is only used for deduplication plugin, it means the `Count` property should not be assigned. Which will be confused to the programmers.
So I think another equality comparer should be created only used for the deduplication specific, which follows the original implementation of the `Event`. Please refer to the commit https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/576782c4f00d1d47fd7ac12a60f178ae0f8ebca1
Another hand, if the `Count` property should not be used by application developers, it should be removed from the `Event` declaration, and implemented with other algorithm in the `DuplicateCheckerPlugin` to avoid confusion.

src/Exceptionless/Serializer/DefaultJsonSerializer.cs
edwardmeng: The original implementation constructed `StreamWriter` will close the `outputStream` in the `Deserialize` method. When the `outputStream` is used to further more manipulation after deserialize/serialize, an error will be thrown. Such as the following code:
```cs
using (var memory = new MemoryStream()) {
       serializer.Serialize(ev, memory);
       memory.Position = 0;
       Assert.Equal(ev, serializer.Deserialize<Event>(memory));
 }
```
In my opinion, the `outputStream` close time should be controlled by the callers, but not the serializer.
Now there is another commit https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/0ad986bcedb2f09f295abc0b4d80a41c24d13b41, there is more parameters passed when constructing `StreamWriter` or `StreamReader` to ensure leave the `outputStream`.

test/Exceptionless.MessagePack.Tests/Exceptionless.MessagePack.Tests.csproj
edwardmeng: @niemyjski I have noticed the `TargetFramework` of the `Exceptionless.Tests` project is `net462`, so the `Exceptionless.MessagePack.Tests` just follows. As your suggestion, we should improve both the projects together.

test/Exceptionless.MessagePack.Tests/Exceptionless.MessagePack.Tests.csproj
edwardmeng: I have removed this reference, please refer to the commit https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/df4cf6d7e1bc7136192c4dadb15aa2194f478ccd

test/Exceptionless.Tests/Serializer/SerializerTests.cs
edwardmeng: @niemyjski The unit test class has been changed as your suggestion, please refer to the commit https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/df7186658cf090771d8285c500e757e9c0236b24

edwardmeng: @niemyjski Please refer to the pr https://github.com/exceptionless/Exceptionless.Net/pull/166, @ejsmith suggested to serialize events and settings through `IStorageSerializer`. 
If you wanna the another interface `ISettingsSerializer` to serialize settings only, we should have further more discussion.

src/Platforms/Exceptionless.MessagePack.Signed/Exceptionless.MessagePack.Signed.csproj
edwardmeng: I think we should create another task to remove all the signed packages once a time. It's up to your plan.

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
niemyjski: I think we should remove this and put this into a different PR where we can discus it more. We should keep this pr focused.  It's potentially a massive change without knowing the exact consequences without lots of testing, dog fooding and benchmarks. To me Equals and GetHashCode should always be checking the same fields

src/Exceptionless/Models/Client/Event.cs
niemyjski: @edwardmeng we control the properties and we know these can change from event to event but the event is the same and we want it stacked. Yes, it shouldn't be set but we know for sure things like dates shouldn't be included. I can see where you are coming from but this is a known limitation that has provides more value than having exact hashcode on every field (but open to discussion). Let's move this to a different PR so it can be discussed.

niemyjski:
Latest changes look good,. Just want to move the equalities changes out to a different pr so we can work on only that and get the message pack features merged in.

niemyjski:
Thanks for the updates, we are very grateful and the latest changes look good. Talking to @ejsmith and @srijken we think we should move the equalities, get hash code changes out to a different pr so we can discus this and work on that separately and keep this PR focused on message pack and get it merged in quickly.

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
ejsmith: I agree @edwardmeng we really need to keep this PR focused on MessagePack and then we can start another one with these changes and have a discussion.

edwardmeng:
The changes about equality compare methods(Equals and GetHashCode) have been reverted, and added another comparer only for unit test purpose, please refer to https://github.com/exceptionless/Exceptionless.Net/pull/168/commits/55d33eaac623383d47a19753a8a54a3c75fcbf7d.
I agree with that we should create new pr to talk about the changes.

edwardmeng:
The unit tests of Exceptionless.MessagePack.Tests were not run in appveyor. Since they are all passed in developing environment through `dotnet xunit`, but all failed in appveyor build. After investigation, I think the root cause is:
The xUnit version is 2.0.0 RTM  that the appveyor is running with, but the xUnit version is 2.3.1 that in projects reference.

edwardmeng:
The Exceptionless.MessagePack.Tests project has been changed to support netcoreapp2.0 and net462, and fixed bug of event deserialization and unit test running issues in appveyor.

src/Exceptionless/Extensions/CollectionEqualityExtensions.cs
```diff
@@ -5,6 +5,18 @@
 namespace Exceptionless {
     internal static class CollectionEqualityExtensions {
         public static bool CollectionEquals<T>(this IEnumerable<T> source, IEnumerable<T> other) {
+            if (ReferenceEquals(null, source) && ReferenceEquals(null, other)) {
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
```diff
@@ -101,7 +101,7 @@ public class EnvironmentInfo : IData {
...
+            return ProcessorCount == other.ProcessorCount && TotalPhysicalMemory == other.TotalPhysicalMemory && AvailablePhysicalMemory == other.AvailablePhysicalMemory && string.Equals(CommandLine, other.CommandLine) && string.Equals(ProcessName, other.ProcessName) && string.Equals(ProcessId, other.ProcessId) && ProcessMemorySize == other.ProcessMemorySize && string.Equals(ThreadName, other.ThreadName) && string.Equals(ThreadId, other.ThreadId) && string.Equals(Architecture, other.Architecture) && string.Equals(OSName, other.OSName) && string.Equals(OSVersion, other.OSVersion) && string.Equals(IpAddress, other.IpAddress) && string.Equals(MachineName, other.MachineName) && string.Equals(InstallId, other.InstallId) && string.Equals(RuntimeVersion, other.RuntimeVersion) && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/InnerError.cs
```diff
@@ -43,7 +43,7 @@ public class InnerError : IData {
         public Method TargetMethod { get; set; }
 
         protected bool Equals(InnerError other) {
-            return string.Equals(Message, other.Message) && string.Equals(Type, other.Type) && string.Equals(Code, other.Code) && Equals(Data, other.Data) && Equals(Inner, other.Inner) && StackTrace.CollectionEquals(other.StackTrace) && Equals(TargetMethod, other.TargetMethod);
+            return string.Equals(Message, other.Message) && string.Equals(Type, other.Type) && string.Equals(Code, other.Code) && Data.CollectionEquals(other.Data) && Equals(Inner, other.Inner) && StackTrace.CollectionEquals(other.StackTrace) && Equals(TargetMethod, other.TargetMethod);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/ManualStackingInfo.cs
```diff
@@ -29,5 +29,25 @@ public class ManualStackingInfo {
         /// Key value pair that determines how the event is stacked.
         /// </summary>
         public IDictionary<string, string> SignatureData { get; set; }
+
+        protected bool Equals(ManualStackingInfo other) {
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/Method.cs
```diff
@@ -20,7 +20,7 @@ public class Method : IData {
         public ParameterCollection Parameters { get; set; }
 
         protected bool Equals(Method other) {
-            return IsSignatureTarget == other.IsSignatureTarget && string.Equals(DeclaringNamespace, other.DeclaringNamespace) && string.Equals(DeclaringType, other.DeclaringType) && string.Equals(Name, other.Name) && Equals(Data, other.Data) && GenericArguments.CollectionEquals(other.GenericArguments) && Parameters.CollectionEquals(other.Parameters);
+            return IsSignatureTarget == other.IsSignatureTarget && string.Equals(DeclaringNamespace, other.DeclaringNamespace) && string.Equals(DeclaringType, other.DeclaringType) && string.Equals(Name, other.Name) && Data.CollectionEquals(other.Data) && GenericArguments.CollectionEquals(other.GenericArguments) && Parameters.CollectionEquals(other.Parameters);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/Module.cs
```diff
@@ -27,7 +27,7 @@ public class Module : IData {
         }
 
         protected bool Equals(Module other) {
-            return ModuleId == other.ModuleId && string.Equals(Name, other.Name) && string.Equals(Version, other.Version) && IsEntry == other.IsEntry && CreatedDate.Equals(other.CreatedDate) && ModifiedDate.Equals(other.ModifiedDate) && Equals(Data, other.Data);
+            return ModuleId == other.ModuleId && string.Equals(Name, other.Name) && string.Equals(Version, other.Version) && IsEntry == other.IsEntry && CreatedDate.Equals(other.CreatedDate) && ModifiedDate.Equals(other.ModifiedDate) && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/Parameter.cs
```diff
@@ -15,7 +15,7 @@ public class Parameter : IData {
         public GenericArguments GenericArguments { get; set; }
 
         protected bool Equals(Parameter other) {
-            return string.Equals(Name, other.Name) && string.Equals(Type, other.Type) && string.Equals(TypeNamespace, other.TypeNamespace) && Equals(Data, other.Data) && GenericArguments.CollectionEquals(other.GenericArguments);
+            return string.Equals(Name, other.Name) && string.Equals(Type, other.Type) && string.Equals(TypeNamespace, other.TypeNamespace) && Data.CollectionEquals(other.Data) && GenericArguments.CollectionEquals(other.GenericArguments);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/RequestInfo.cs
```diff
@@ -70,7 +70,7 @@ public class RequestInfo : IData {
         public DataDictionary Data { get; set; }
 
         protected bool Equals(RequestInfo other) {
-            return string.Equals(UserAgent, other.UserAgent) && string.Equals(HttpMethod, other.HttpMethod) && IsSecure == other.IsSecure && string.Equals(Host, other.Host) && Port == other.Port && string.Equals(Path, other.Path) && string.Equals(Referrer, other.Referrer) && string.Equals(ClientIpAddress, other.ClientIpAddress) && Cookies.CollectionEquals(other.Cookies) && QueryString.CollectionEquals(other.QueryString) && Equals(Data, other.Data);
+            return string.Equals(UserAgent, other.UserAgent) && string.Equals(HttpMethod, other.HttpMethod) && IsSecure == other.IsSecure && string.Equals(Host, other.Host) && Port == other.Port && string.Equals(Path, other.Path) && string.Equals(Referrer, other.Referrer) && string.Equals(ClientIpAddress, other.ClientIpAddress) && Cookies.CollectionEquals(other.Cookies) && QueryString.CollectionEquals(other.QueryString) && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/SimpleError.cs
```diff
@@ -32,7 +32,7 @@ public class SimpleError : IData {
         public SimpleError Inner { get; set; }
 
         protected bool Equals(SimpleError other) {
-            return string.Equals(Message, other.Message) && string.Equals(Type, other.Type) && string.Equals(StackTrace, other.StackTrace) && Equals(Data, other.Data) && Equals(Inner, other.Inner);
+            return string.Equals(Message, other.Message) && string.Equals(Type, other.Type) && string.Equals(StackTrace, other.StackTrace) && Data.CollectionEquals(other.Data) && Equals(Inner, other.Inner);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/StackFrame.cs
```diff
@@ -7,7 +7,7 @@ public class StackFrame : Method {
         public int Column { get; set; }
 
         protected bool Equals(StackFrame other) {
-            return base.Equals(other) && string.Equals(FileName, other.FileName);
+            return base.Equals(other) && string.Equals(FileName, other.FileName) && LineNumber == other.LineNumber && Column == other.Column;
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/StackFrame.cs
```diff
@@ -24,6 +24,8 @@ public class StackFrame : Method {
             unchecked {
                 int hashCode = base.GetHashCode();
                 hashCode = (hashCode * 397) ^ (FileName == null ? 0 : FileName.GetHashCode());
+                hashCode = (hashCode * 397) ^ LineNumber.GetHashCode();
+                hashCode = (hashCode * 397) ^ Column.GetHashCode();
```
niemyjski: This needs to be reverted and can't go into the new pr for equality changes. Line numbers and columns can change at random (via runtime...). It was difficult to track this down but we excluded it for these reasons.

src/Exceptionless/Models/Client/Data/UserDescription.cs
```diff
@@ -23,7 +23,7 @@ public class UserDescription : IData {
         public DataDictionary Data { get; set; }
 
         protected bool Equals(UserDescription other) {
-            return string.Equals(EmailAddress, other.EmailAddress) && string.Equals(Description, other.Description) && Equals(Data, other.Data);
+            return string.Equals(EmailAddress, other.EmailAddress) && string.Equals(Description, other.Description) && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Data/UserInfo.cs
```diff
@@ -32,7 +32,7 @@ public class UserInfo : IData {
         public DataDictionary Data { get; set; }
 
         protected bool Equals(UserInfo other) {
-            return string.Equals(Identity, other.Identity) && string.Equals(Name, other.Name) && Equals(Data, other.Data);
+            return string.Equals(Identity, other.Identity) && string.Equals(Name, other.Name) && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

src/Exceptionless/Models/Client/Event.cs
```diff
@@ -59,7 +59,7 @@ public class Event : IData {
         public string ReferenceId { get; set; }
 
         protected bool Equals(Event other) {
-            return string.Equals(Type, other.Type) && string.Equals(Source, other.Source) && Tags.CollectionEquals(other.Tags) && string.Equals(Message, other.Message) && string.Equals(Geo, other.Geo) && Value == other.Value && Equals(Data, other.Data);
+            return string.Equals(Type, other.Type) && string.Equals(Source, other.Source) && Tags.CollectionEquals(other.Tags) && string.Equals(Message, other.Message) && string.Equals(Geo, other.Geo) && Value == other.Value && Data.CollectionEquals(other.Data);
```
niemyjski: Can we revert all equality changes and move them into a different pr

test/Exceptionless.TestHarness/Serializer/EventEqualityComparer.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Collections.Generic;
+using Exceptionless.Models;
+
+namespace Exceptionless.Tests.Serializer {
+    public class EventEqualityComparer : IEqualityComparer<Event> {
```
niemyjski: Can we revert all equality changes and move them into a different pr

test/Exceptionless.Tests/Plugins/PluginTests.cs
```diff
@@ -856,6 +856,69 @@ public class CounterTestPlugin : IEventPlugin, IDisposable {
             }
         }
 
+        [Fact(Skip = "The duplicate checker verify events with hash code is not accurate.")]
```
niemyjski: Can we revert all equality changes and move them into a different pr

test/Exceptionless.Tests/Serializer/JsonStorageSerializerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        protected override void Initialize(IDependencyResolver resolver) {
+            base.Initialize(resolver);
+            resolver.Register<IStorageSerializer, DefaultJsonSerializer>();
+        }
+
+        protected override IStorageSerializer GetSerializer(IDependencyResolver resolver) {
+            return resolver.Resolve<IStorageSerializer>();
+        }
+
+        [Fact(Skip = "The json serializer deserialize anonymous(object) types as dictionary.")]
```
niemyjski: Should these all pass if we are using the same serializer (the default one?)

edwardmeng: none of them will pass based on the reasons:
1. Incorrect equality compare algorithm
2. The value deserialized in data dictionary are type of Dictionary<string,object> but not the UserDescription, UserInfo or EnvironmentInfo etc.

edwardmeng:
The equality compare relative methods have been reverted as your suggestion, and the unit tests for message pack serializer are skipped. The unit tests for duplication checker have been removed too.

niemyjski:
Thanks a million for the help, pr and reverting the equality changes (Helps keep it focused and easier to test / get in). Your help is greatly appreciated. Do you need us to push a release out there soon for this or do you want the equality changes in with the next release?

From looking on your tests and talking with you and @srijken  there looks like there are some equality issues that need to be addressed. Here are our concerns (we should open a new pr and or issue to discus but I just wanted to put them out there :). 

1. We need more benchmarks of previous GetHashCode implementation + equals implementation on how it effects the plugin pipeline / duplicate checker (we have a few duplicate checker plugins). This is an extremely hot code path for everyone and we need to ensure we do not make it slower at all costs.
2. I know that we cannot add Count, Date (there was another one) event properties to GetHashCode() as those can change from event to event and those properties to us don't determine equality. On the previous comment, we've determined overtime that LineNumber and LineColumn cannot be included. These two values can and will change during runtime and will break deduplication. Basically the JIT can change the offset as well as the state of symbols (if I remember correctly).

On a side note, we made you a contributor a week ago :). Please feel free to join us on slack.exceptionless.com :)

edwardmeng:
I recommend the next release to include the equality changes and duplicate checker improvements. The following days I will busy again, and have no time to request pr. I am so sorry for that.

zhouyongtao:
It's looks like a pretty cool

PR #185, opened 2018-04-17
Added request content logging for Exceptionless.WebApi

PolitovArtyom:
Adapted solution from  [Exceptionless.AspNetCore ](https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs)

But really, i don't like it. 
I think we should add settings for enabling this feature and setting log message max content size. Otherwise, changes may affect users with large requests payload.

### DIFF
--- a/src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
+++ b/src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
@@ -12,6 +13,18 @@ using Exceptionless.Models.Data;
 
 namespace Exceptionless.ExtendedData {
     internal static class RequestInfoCollector {
+        private const long MAX_CONTENT_LENGTH = 1024 * 50;
+
+        private static readonly List<string> _ignoredFormFields = new List<string> {
+            "__*"
+        };
+
+        private static readonly List<string> _ignoredCookies = new List<string> {
+            ".ASPX*",
+            "__*",
+            "*SessionId*"
+        };
+
         public static RequestInfo Collect(HttpActionContext context, ExceptionlessConfiguration config) {
             if (context == null)
                 return null;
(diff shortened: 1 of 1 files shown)


src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
```diff
@@ -104,9 +108,45 @@ internal static class RequestInfoCollector {
 
                 if (request.Properties.ContainsKey(RemoteEndpointMessageProperty.Name))
                     return ((RemoteEndpointMessageProperty)request.Properties[RemoteEndpointMessageProperty.Name]).Address;
-            } catch {}
+            }
+            catch { }
```
niemyjski: Can we please revert the styling changes :)

src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
```diff
@@ -78,7 +81,8 @@ internal static class RequestInfoCollector {
                 try {
                     string value = values.Get(key);
                     d.Add(key, value);
-                } catch (Exception ex) {
+                }
+                catch (Exception ex) {
```
niemyjski: Can we please revert the styling changes :)

src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
```diff
@@ -104,9 +108,45 @@ internal static class RequestInfoCollector {
...
                     return ((RemoteEndpointMessageProperty)request.Properties[RemoteEndpointMessageProperty.Name]).Address;
-            } catch {}
+            }
+            catch { }
 
             return String.Empty;
         }
+
+        private static object GetPostData(HttpContent httpContent, string[] exclusionList) {
+            if (httpContent.IsFormData()) {
```
niemyjski: We need to add a null check here for http content and return just to be safe

src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
```diff
@@ -104,9 +108,45 @@ internal static class RequestInfoCollector {
...
+                var content = httpContent.ReadAsFormDataAsync().GetAwaiter().GetResult();
```
niemyjski: Is there a non async method to do this? It kinda stinks to be doing a GetResult here and potentially blocking the app. Also we need to test what occurs if you try to log in the start of a middleware, I recently got a report and need to look into it that it could break the controller bindings because the stream is read.

src/Platforms/Exceptionless.WebApi/RequestInfoCollector.cs
```diff
@@ -104,9 +108,45 @@ internal static class RequestInfoCollector {
...
+                return GetContent(httpContent);
+            }
+
+            var contentSize = Math.Round(httpContent.Headers.ContentLength.Value / 1024m, 0).ToString("N0");
+            return $"Data is too large ({contentSize}kb) to be included.";
+        }
+
+        private static object GetContent(HttpContent httpContent) {
+            try {
+                var contentStream = httpContent.ReadAsStreamAsync().GetAwaiter().GetResult();
```
niemyjski: I need to look at our asp.net core implementation but I think we look at the stream length to see if it's too long after we have read it, otherwise it could be a massive post body and the client will get rejected when it's submitted.

PolitovArtyom:
I've researched problems with synchronus invocation of 'httpContent.ReadAsFormDataAsync()' and  'HttpContent' flush afrer first read.  
And there are no solutions.  Logging request body in ASP.NET WebApi is potentioally dangerous for external code. So i'm going to close this merge request.

For my case i've implemented plugin with request body logging. 
If you think plugn may be included in Exceptionless.WebApi, i can create new merge request.

PR #186, opened 2018-04-18
Exceptionless.Mvc added logging with custom ExceptionlessClient

PolitovArtyom:
Added logging with custom ExceptionlessClient to Exceptionless.Mvc.

### DIFF
--- a/src/Platforms/Exceptionless.Mvc/ExceptionlessModule.cs
+++ b/src/Platforms/Exceptionless.Mvc/ExceptionlessModule.cs
@@ -9,24 +9,33 @@ using Exceptionless.Web.Extensions;
 
 namespace Exceptionless.Mvc {
     public class ExceptionlessModule : IHttpModule {
+        private readonly ExceptionlessClient _client;
         private HttpApplication _app;
 
+        public ExceptionlessModule() {
+            _client = ExceptionlessClient.Default;
+        }
+
+        public ExceptionlessModule(ExceptionlessClient client = null) {
+            _client = client ?? throw new ArgumentNullException();
+        }
+
         public virtual void Init(HttpApplication app) {
-            ExceptionlessClient.Default.Startup();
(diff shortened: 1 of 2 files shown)


PolitovArtyom:
I need to log all errors in my MVC application, so i don't need `ExceptionlessSendErrorsAttribute`'at all.
I can configure client like this:
```
_client.RegisterHttpApplicationErrorHandler(app);
_client.Configuration.AddPlugin<ExceptionlessWebPlugin>();
```
But `ExceptionlessWebPlugin` is internal. Is there are any reasons for it?

Maybe best solution for my problem is just to change `ExceptionlessWebPlugin`  access rights to public

niemyjski:
We recommend that you just add the http module to the web.config otherwise you won't be able to capture all exceptions in your app as the module can fire before your application is initialized. We try and keep the internals private unless a good argument can be made to make it public. Generally they are not extended. Can you not use an HTTP Module?

PolitovArtyom:
Usage of web.config file for my case will require updating web.config for 6 APIs and 3 sets of configuration (dev, test, prod). I don't think it's a good approach. For deploy purposes my application configuration combined in single file, 

All i need is a bit more flexibility for exceptionless configuration via code. 
As i see 'ExceptionlessWebPlugin'  is a main feature of Exceptionless.Mvc package.  'ExceptionlessModule' and 'ExceptionlessSendErrorsAttribute' is just a helpers for logging registration.

niemyjski:
I guess we could make it public. Have you thought about just wiring up the module only in the web.config? Then we'd automatically pick up the api key from environment variables or app settings (as defined in azure app properties). You wouldn't have to keep changing the config between environments.

PolitovArtyom:
It's a good solution for my case.  I was confused by usage of global property ExceptionlessClient.Default. But for logging purposes it should be ok.

Thank your for listening!

PR #187, opened 2018-05-01
GPDR Changes

niemyjski:
References https://github.com/exceptionless/Exceptionless/issues/370

- [x] Add individual flags to turn on or off specific functionality. This may also improve performance of collecting of Cookies / PostData etc..
- [x] Discus and implement configuration changes to quickly toggle these settings.
- [x] Add Wiki Page and Update Readme to talk about configuration.

We need to discus whether IncludePrivateData needs to always set some of these flags as some of these values like Cookies / Form Data (excluding post data) / QueryString already get filtered via data exclusions.

Also, Should we add a plugin that Hashes UserName and or Computer Name? So sessions always work.


ejsmith:
The settings should be defaulted to true to match existing behavior.

PR #189, opened 2018-05-04
Introduce Android.PreserveAttribute to constructors removed by Xamarin Linker

### LINKED ISSUE #188: Exceptionless will fail to initialize due to TinyIoC resolution exception in Xamarin Android with linking enabled
Due to the usage of dependency injection container instead of calling constructors directly, Xamarin's linker will remove constructors for a lot of classes and Exceptionless will fail to start.

A trivial reproduction can be found here: https://github.com/GeirGrusom/ExceptionlessException

GeirGrusom:
This pull-request aims to make sure that the Xamarin Linker does not remove constructors that are used only by the DI container, which the linker cannot know.

It makes no functional changes to the code, but should allow Xamarin applications to link Exceptionless using the default setup.

Fixes #188


niemyjski:
Can we rename the namespace to something more generic (doesn't include a platform). I'm find with renaming the attribute to to something that describes it better (PreserveFromTreeShaking?) but not sure on a name. What all configuration is required for xamarin to look at these attributes?

GeirGrusom:
Namespace can be changed, but the name of the type must remain. How about `Exceptionless.Linking`? Or do you have a preference?

> just wonder if someone would know why it's there and what classes it's required on

I'll add documentation on the type to make it clear what it's for. This type is `internal` so it won't get exposed to library users directly.

> What all configuration is required for xamarin to look at these attributes?

If you enable linking it will look for an attribute named `PreserveAttribute`. It doesn't matter what namespace it belongs to. 

This is from the documentation:

> It does not really matter in which namespace this is defined, the linker looks this attribute by type name.

You could exclude the assmbly as well, but I couldn't get that to work for some reason. It just ignored it. It's also a pita if everyone has to step in this for every Xamarin project.

So what I'm afraid of with this is that it may make Exceptionless flimsy in Xamarin solutions. Exceptionless isn't linker safe, so there may still be cases where the linker ends up removing constructors from types used in certain configurations since I haven't marked all types, or the entire assembly, with `[Preserve]`. This fix seems to resolve my issue, so at least Exceptionless starts up and manages to send me exception messages. I've preserved the constructors on all plugins.

GeirGrusom:
> any reason for not marking the whole assembly?

No. I think that would perhaps be the safest thing to do. The disadvantage is that the linker won't be able to trim the assembly at all, but that's a small price to pay I would think.

I'll change the namespace and apply the attribute to the assembly instead.

niemyjski:
Looks great to me! Thanks for the PR

PR #190, opened 2018-06-04
POST data can get lost

srijken:


src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +46,70 @@ internal static class RequestInfoCollector {
...
+        private static object GetPostData(HttpContext context, ExceptionlessConfiguration config, string[] exclusionList) {
+            var log = config.Resolver.GetLog();
+
+            if (context.Request.HasFormContentType && context.Request.Form.Count > 0) {
+                log.Debug("Reading POST data from Request.Form");
+
+                return context.Request.Form.ToDictionary(exclusionList);
+            }
+
+            if (!context.Request.ContentLength.HasValue || context.Request.ContentLength.Value == 0)
```
niemyjski: We should add a log statement here because some http clients don't set the content length.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +46,70 @@ internal static class RequestInfoCollector {
...
+                string message = String.Format("Data is too large ({0}kb) to be included.", value);
+                log.Debug(message);
+                return message;
+            }
+
+            try {
+                if (!context.Request.Body.CanSeek) {
+                    string message = "Unable to get POST data: The stream could not be reset.";
+                    log.Debug(message);
+
```
niemyjski: Extra whitespace between logs and returns :)

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +46,70 @@ internal static class RequestInfoCollector {
...
+            if (context.Request.ContentLength.Value >= 1024 * 50) {
+                string value = Math.Round(context.Request.ContentLength.Value / 1024m, 0).ToString("N0");
+                string message = String.Format("Data is too large ({0}kb) to be included.", value);
+                log.Debug(message);
+                return message;
+            }
+
+            try {
+                if (!context.Request.Body.CanSeek) {
```
niemyjski: Before we were reading the stream as long as the position was 0, now we are only reading it if we can seek it and return it back. I think I'm fine with this change but it sucks if the request failed in the middleware, then it doesn't really matter. I wonder if we could be smart and see if the response was even generated?

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +46,70 @@ internal static class RequestInfoCollector {
...
+                if (context.Request.Body.Position != 0) {
+                    string message = "Unable to get POST data: The stream position was not at 0.";
+                    log.Debug(message);
+
+                    return message;
+                }
+
+                // pass default values, except for leaveOpen: true. This prevents us from disposing the underlying stream
+                using (var inputStream = new StreamReader(context.Request.Body, Encoding.UTF8, true, 1024, true)) {
+                    string postData = inputStream.ReadToEnd();
```
niemyjski: Since content length headers can lie. I wonder if we should be doing some kind of length check here as well that we weren't doing previously.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
srijken: Yeah but reading it when it's at 0 and you can't set it back to 0 is a pain, because that will also cause POST data to get lost. This is different behavior than before, but this is also the essence of the fix I think (apart from disposing the body stream)

so if CanSeek is false, and Position is 0, it's pretty much guaranteed to cause problems later on (otherwise the body would have been read anyway)

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
srijken: for what case is this? When there's more than 50kB, but the content length indicates less than 50kB?

niemyjski: That too, but if the client doesn't set the content length this will be null. I believe one manually has to set this if you created your own httpclient instance..

srijken: So we'll have to adjust L63-64 then, because we already bail out when that happens. Not sure what happens with HttpClient.Post/PutAsync

niemyjski: We don't have to, but it might be nice to. I don't know what the overhead would be to do that.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +47,71 @@ internal static class RequestInfoCollector {
...
+        private static object GetPostData(HttpContext context, ExceptionlessConfiguration config, string[] exclusionList) {
+            var log = config.Resolver.GetLog();
+
+            if (context.Request.HasFormContentType && context.Request.Form.Count > 0) {
+                log.Debug("Reading POST data from Request.Form");
+                return context.Request.Form.ToDictionary(exclusionList);
+            }
+
+            var maxDataToRead = context.Request.ContentLength == 0 ? MAX_BODY_SIZE : context.Request.ContentLength;
+            if (maxDataToRead > 1024 * 50) {
```
niemyjski: Probably should replace this with the constant `MAX_BODY_SIZE`

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +47,71 @@ internal static class RequestInfoCollector {
...
 
+        private static object GetPostData(HttpContext context, ExceptionlessConfiguration config, string[] exclusionList) {
+            var log = config.Resolver.GetLog();
+
+            if (context.Request.HasFormContentType && context.Request.Form.Count > 0) {
+                log.Debug("Reading POST data from Request.Form");
+                return context.Request.Form.ToDictionary(exclusionList);
+            }
+
+            var maxDataToRead = context.Request.ContentLength == 0 ? MAX_BODY_SIZE : context.Request.ContentLength;
```
niemyjski: Be better to store this as a variable context.Request.ContentLength.GetValueOrDefault() and use it everyplace where content length is used.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +47,71 @@ internal static class RequestInfoCollector {
...
+                    log.Debug(message);
+                    return message;
+                }
+
+                // pass default values, except for leaveOpen: true. This prevents us from disposing the underlying stream
+                using (var inputStream = new StreamReader(context.Request.Body, Encoding.UTF8, true, 1024, true)) {
+                    var sb = new StringBuilder();
+                    int numRead;
+                    char[] buffer = new char[1024];
+                    while ((numRead = inputStream.ReadBlock(buffer, 0, 1024)) > 0 && (sb.Length + numRead) < maxDataToRead) {
```
niemyjski: What happens for binary content or the content length is less than the buffer? We are defensive against a wrongly reported content length size but perhaps we should adjust the buffer to be the reported content size if it's smaller so we only do one read and if it's wrongly reported we bail.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -44,33 +47,71 @@ internal static class RequestInfoCollector {
...
+                        sb.Append(buffer, 0, numRead);
+                    }
+                    string postData = sb.ToString();
+
+                    context.Request.Body.Position = originalPosition;
+
+                    log.Debug($"Reading POST, set back to position: {context.Request.Body.Position}");
```
niemyjski: We could use originalPosition here and save some accessor reads.

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
srijken: no with the variable that contains the contentLenght. `MAX_BODY_SIZE` will never be greater than 50k :)

PR #194, opened 2018-07-31
Server certificate validation

### LINKED ISSUE #192: Certificate is not validated in NET45
https://github.com/exceptionless/Exceptionless.Net/blob/0ee9d35dce01105891910fe9857f584a4c813469/src/Exceptionless/Submission/DefaultSubmissionClient.cs#L123-L127

We've relied on this behavior since our own internal cert is not trusted on all machines, but we still think it's wrong to open all NET45 applications to an MiM Attack. We found this while migrating to ASP.NET Core which is always validating the certificate.

I would suggest to add a validation handler to the configuration to allow a simple validation for all existing users without creating a custom `ISubmissionClient`.

zivillian:
fixes #192

### DIFF
--- a/src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
+++ b/src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
@@ -15,6 +15,7 @@ using Exceptionless.Storage;
 
 #if !PORTABLE && !NETSTANDARD1_2
 using Exceptionless.Diagnostics;
+using System.Net.Security;
 #endif
 
 #if NET45
@@ -432,6 +433,46 @@ namespace Exceptionless {
                 return null;
 
             return _environmentVariables[name];
+        }
+#endif
+
+#if !PORTABLE && !NETSTANDARD1_2
+        /// <summary>
+        /// Add a custom server certificate validation against the thumbprint of the server certificate.
+        /// </summary>
+        /// <param name="config">The configuration object you want to apply the attribute settings to.</param>
+        /// <param name="thumbprint">Thumbprint of the server certificate. <example>e.g. "86481791CDAF6D7A02BEE9A649EA9F84DE84D22C"</example></param>
+        public static void TrustCertificateThumbprint(this ExceptionlessConfiguration config, string thumbprint) {
+            config.ServerCertificateValidationCallback = x => {
+                if (x.SslPolicyErrors == SslPolicyErrors.None) return true;
+                return x.Certificate != null && thumbprint != null && thumbprint.Equals(x.Certificate.Thumbprint, StringComparison.OrdinalIgnoreCase);
+            };
+        }
+
+        /// <summary>
+        /// Add a custom server certificate validation against the thumbprint of any of the ca certificates.
+        /// </summary>
+        /// <param name="config">The configuration object you want to apply the attribute settings to.</param>
+        /// <param name="thumbprint">Thumbprint of the ca certificate. <example>e.g. "afe5d244a8d1194230ff479fe2f897bbcd7a8cb4"</example></param>
+        public static void TrustCAThumbprint(this ExceptionlessConfiguration config, string thumbprint) {
+            config.ServerCertificateValidationCallback = x => {
+                if (x.SslPolicyErrors == SslPolicyErrors.None) return true;
+                if (x.Chain == null || thumbprint == null) return false;
+                foreach (var ca in x.Chain.ChainElements) {
+                    if (thumbprint.Equals(ca.Certificate.Thumbprint, StringComparison.OrdinalIgnoreCase))
+                        return true;
+                }
+                return false;
+            };
+        }
+
+        /// <summary>
+        /// Disable any certificate validation. Do not use this in production.
+        /// </summary>
+        /// <param name="config"></param>
+        [Obsolete("This will open the client to Man-in-Middle attacks. It should never be used in production.")]
+        public static void SkipCertificateValidation(this ExceptionlessConfiguration config) {
+            config.ServerCertificateValidationCallback = x => true;
         }
 #endif
 
--- a/src/Exceptionless/Submission/DefaultSubmissionClient.cs
+++ b/src/Exceptionless/Submission/DefaultSubmissionClient.cs
@@ -4,6 +4,10 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+#if NET45 || (!PORTABLE && !NETSTANDARD1_2)
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+#endif
 using System.Text;
 using Exceptionless.Configuration;
 using Exceptionless.Dependency;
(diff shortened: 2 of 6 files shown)


src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
```diff
@@ -435,6 +436,46 @@ public static class ExceptionlessConfigurationExtensions {
...
+                }
+                return false;
+            };
+        }
+
+        /// <summary>
+        /// Disable any certificate validation. Do not use this in production.
+        /// </summary>
+        /// <param name="config"></param>
+        [Obsolete("This will open the client to Man-in-Middle attacks. It should never be used in production.")]
```
niemyjski: I'm not sure this should be obsolete, I think this still needs to be the default behavior and we update the configuration guide on how to better handle this.

src/Exceptionless/Submission/DefaultSubmissionClient.cs
```diff
@@ -123,12 +127,28 @@ public class DefaultSubmissionClient : ISubmissionClient, IDisposable {
...
             var handler = new WebRequestHandler { UseDefaultCredentials = true };
-            handler.ServerCertificateValidationCallback = delegate { return true; };
 #else
             var handler = new HttpClientHandler { UseDefaultCredentials = true };
+#endif
 #if !PORTABLE && !NETSTANDARD1_2
-            //handler.ServerCertificateCustomValidationCallback = delegate { return true; };
+            var callback = config.ServerCertificateValidationCallback;
+            if (callback != null) {
+#if NET45
```
niemyjski: Can we make this into dedicated methods. I know local functions are compiler sugar but it can cause runtime issues on say unmatched windows 7 machines (at least that's what I've found with other c# new features).

src/Exceptionless/Submission/DefaultSubmissionClient.cs
```diff
@@ -123,12 +127,28 @@ public class DefaultSubmissionClient : ISubmissionClient, IDisposable {
...
+                }
+
+                handler.ServerCertificateValidationCallback = Validate;
+#else
+                bool Validate(HttpRequestMessage httpRequestMessage, X509Certificate2 certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors) {
+                    var certData = new CertificateData(httpRequestMessage, certificate, chain, sslPolicyErrors);
+                    return callback(certData);
+                }
+
+                handler.ServerCertificateCustomValidationCallback = Validate;
```
niemyjski: We have to be very careful here. Can you see if the supported flag was added here (https://github.com/dotnet/corefx/issues/9728). Please see this thread for more information on why this has been dangerous on non full framework.

zivillian: This should be fixed with 2.1 as curl was replaced by SocketsHttpHandler (https://blogs.msdn.microsoft.com/dotnet/2018/04/11/announcing-net-core-2-1-preview-2/#user-content-sockets-performance-and-socketshttphandler). I'm unsure how to detect this.

niemyjski: Our clients are compiled for .net Standard 1.2 and above, we can't guarantee that our clients are running 2.1 and need to be safe.

zivillian: The handler will only be installed if the user used one of the extension methods or implemented his own callback. Since this is currently also not supported for netcore version relying on curl this will not break anything.
I guess it should be mentioned in the release notes that there may be a problem with mac or linux on older netcore versions, but I think there is nothing you or I can do about this bug.
The previous implementation also required valid certificates on curl based netcore, so the behaviour is the same.

src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
zivillian: To be honest, I will not change this. I just don't want my name next to an insecure default. If someone needs to troubleshoot the integration, there is this default implementation for things like testing or just getting it to run locally, but this __must__ not be used in production without even thinking about the consequences.
If some user really has the requirement and a good justification he may use this method and supress the warning, or just implement his own version of `I don't care`.
I know this will be a breaking change and should be mentioned clearly in the release notes.

ejsmith: We literally DO NOT have a choice in the matter. We can't use this in the default client because the cert will eventually expire and all deployed clients will all of a sudden not be able to submit events. It is not an option for us to go and tell users to update all of their software that makes use of the Exceptionless client whenever out SSL cert expires.

zivillian: All clients using exceptionless.io have no problem as long as exceptionless.io has a valid certificate (which it should).
All of these extension methods are more or less dirty workarounds for users who a hosting their own exceptionless __without__ a valid (or otherwise trusted) certificate.
Since exceptions may always contain sensitive and security critical data, I think running your own exceptionless instance without a proper sense of security should __not__ be supported in any way.

ejsmith: Ahh. I guess I was thinking about the thumbprint setting. The default would do normal trusted root certificate validation and this is just for people to use a self-signed cert. Sorry for the confusion.

src/Exceptionless/Submission/DefaultSubmissionClient.cs
```diff
@@ -167,6 +157,20 @@ public class DefaultSubmissionClient : ISubmissionClient, IDisposable {
             return client;
         }
 
+#if !PORTABLE && !NETSTANDARD1_2
+#if NET45
```
niemyjski: Let's just do this for net45 and NETSTANDARD 2_0 as <2.0 might be on core 1.1.

niemyjski:
Looks good, lets only include these changes on NET45 and .NET Standard 2.0

src/Exceptionless/Submission/DefaultSubmissionClient.cs
zivillian: Can you tell me why? What is so special about core 1.1, that this feature should not be available? 
The [API](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=netcore-1.1) is available from netstandard 1.3, supported since core 1.0 and hasn't changed since.

niemyjski: Because it's part of the api in .net core 1.1 but broken in the runtime and this will cause all kinds of issues on non windows machines. We need to only include this on full framework or net standard 2.0 builds. If you could make this change that would be greatly appreciated and we could continue with merging this change in. I'm starting to work on testing for the next release and this could still make it in.

niemyjski:
Thanks for the PR

niemyjski:
were going to update our requirements on .net core 1.0 on plat and accept this change as is. Thanks for the pr and work for us. v4.3.1 is going to roll out shortly.

Issue #165, opened 2017-10-18
Make Exceptionless.Json private or embed shared code as DLL resources in binaries


leidegre:
I get why you'd want to lock down what versions of what is being used, not pollute solutions with unnecessary dependencies and I admire the effort to do so. But I have a vanity request/suggestion.

It might be useful to you it might not be but I thought I'd put it up for discussion, I can help with this as well if you think it's a reasonable idea.

I'm annoyed a little bit by the `Exceptionless.Json` namespace as it pops up every now and then in our solutions. I might have misunderstood but there's no reason for any non-exceptionless code to use that namespace, yes?

I have a couple of suggestions.

- I could write a tool that would change the default accessibility for all types in a specific namespace to internal. InternalsVisisbleTo can be used to then share these types within the exceptionless ecosystem.

- Build shared exceptionless code as a separate DLL and embed this DLL as a resource, then load this resource from the assembly manifest when needed. It can be done with just a few lines of code and doesn't require fancy post-build steps. _As a bonus, you can also opt to compress DLLs for a smaller footprint._

With what I've laid out so far, is this something you might consider changing?

niemyjski:
We can't make it private because then you can't use attributes to do exclusions for serialization (We could probably introduce an interface of our own). Perhaps we can make some more of it internal. 

How are you adding exceptionless as a dependency? We only add it to the entry points of our applications so it's not referenced by any core libraries.

@ejsmith thoughts?

ejsmith:
I’ve run into the same issues where I accidentally use the wrong namespace for Json stuff. I think we could probably make most of the classes internal and then maybe even change the name of some of the attributes for doing things like JsonIgnore. It would be a breaking change, but it may be worth it.

leidegre:
It would be a welcome change, it's a minor inconvenience, as you say it should only show up in entry points but every now and then it crops up as an accidental mistake. Additionally, a tooling like ReSharper will pick up the type in these DLLs and offer it as a suggestion from time to time.

ejsmith:
@leidegre would you be able to make this change and submit a PR?

ejsmith:
I think we could actually internalize the whole thing and then just tell people to either just use `DataMember` or `DataMemberIgnore` attributes to control simple name changes and ignoring members or to use their own serializer.

leidegre:
[...]

PR #199, opened 2018-10-17
Make json.net internal

yang-xiaodong:
Hello @ejsmith  @niemyjski , 

I think we have all realized the problems(#165)  about that the Json.NET is `public`. I also encountered these problems in my project. This PR is used to fix this problem.

This PR contains the following modifications:

* Change Newtonsoft.Json to internal
* Rename `JsonIgnoreAttribute` to `DataMemberIgnoreAttribute`
* Make `DataMemberIgnoreAttribute` access level to `public`

Related issues :  #165

### DIFF
--- a/src/Exceptionless/EventBuilder.cs
+++ b/src/Exceptionless/EventBuilder.cs
@@ -115,7 +115,7 @@ namespace Exceptionless {
 
         /// <summary>
         /// Sets an extended property value to include with the event. Use either <paramref name="excludedPropertyNames" /> or
-        /// <see cref="Exceptionless.Json.JsonIgnoreAttribute" /> to exclude data from being included in the event report.
+        /// <see cref="Exceptionless.Json.DataMemberIgnoreAttribute" /> to exclude data from being included in the event report.
         /// </summary>
         /// <param name="name">The name of the object to add.</param>
         /// <param name="value">The data object to add.</param>
@@ -131,7 +131,7 @@ namespace Exceptionless {
 
         /// <summary>
         /// Adds the object to extended data. Use either <paramref name="excludedPropertyNames" /> or
-        /// <see cref="Exceptionless.Json.JsonIgnoreAttribute" /> to exclude data from being included in the event.
+        /// <see cref="Exceptionless.Json.DataMemberIgnoreAttribute" /> to exclude data from being included in the event.
         /// </summary>
         /// <param name="data">The data object to add.</param>
         /// <param name="name">The name of the object to add.</param>
--- a/src/Exceptionless/Newtonsoft.Json/Bson/BsonObjectId.cs
+++ b/src/Exceptionless/Newtonsoft.Json/Bson/BsonObjectId.cs
@@ -31,7 +31,7 @@ namespace Exceptionless.Json.Bson
     /// <summary>
     /// Represents a BSON Oid (object id).
     /// </summary>
-    public class BsonObjectId
+    internal class BsonObjectId
     {
         /// <summary>
         /// Gets or sets the value of the Oid.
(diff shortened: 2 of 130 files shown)


niemyjski:
Hello,

I really appreciate these changes, can you please update the powershell script (https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Exceptionless/update-json.ps1) to internalize it and rename this attribute. Without doing so makes it very very hard for us to upgrade son.net going forward.

ejsmith:
This is a breaking change for anyone that makes use of those attributes. There is another Json attribute that lets you rename a member.

Why would we change the name of JsonIgnoreAttribute to DataMemberIgnore? DataMemberIgnore is a built in framework attribute name that we would be conflicting with. It seems like we could just rely on the fact that JsonNet honors the built in framework DataMember* attributes and just keep everything internal. People would need to use those attributes to control serialization. Only problem with that is that they may want to ignore something they are sending to exceptionless but not ignore it within their app.

yang-xiaodong:
@niemyjski I am manually updating these files, I am a PowerShell noob, it is difficult for me to change the  script.

yang-xiaodong:
@ejsmith 
Sorry, I saw the suggestion of DataMemberIgnore in this [issue comment](https://github.com/exceptionless/Exceptionless.Net/issues/165#issuecomment-340024511). English is not my first language may be an understanding error. Do I need to change it back to `JsonIgnoreAttribute` and set it to `public` ?

niemyjski:
Hello, yes that would be preferred just to make that public, also can we update the powershell script to make these changes so we can upgrade easily to newer versions of json.net .

ejsmith:
I was thinking that we should rename them to ExceptionlessIgnoreAttribute. So you know that you are ignoring just for Exceptionless.

yang-xiaodong:
@niemyjski Can I use Python script to replace PowerShell script?

niemyjski:
We'd prefer to keep it powershell due to it runs cross plat also it already exists, we'd just need to update/add 2-3 replacement statements.

yang-xiaodong:
@niemyjski 
I have updated the PowerShell script, and test passed on macOS.

src/Exceptionless/update-json.ps1
```diff
@@ -4,31 +4,36 @@ if (Test-Path json.zip) {
...
 
-Get-ChildItem '.\Newtonsoft.Json' *.cs -recurse |
+Get-ChildItem './Newtonsoft.Json' *.cs -recurse |
     Foreach-Object {
         $c = ($_ | Get-Content) 
         $c = $c -replace 'Newtonsoft.Json','Exceptionless.Json'
+        if($_.name -ne 'JsonIgnoreAttribute.cs'){
+            $c = $c -replace 'public( (?:static|sealed|abstract))? (class|struct|interface|enum)','internal$1 $2'
+            $c = $c -replace 'public delegate void','internal delegate void'
+            $c = $c -replace '\[CLSCompliant\(false\)\]',''
```
niemyjski: Any reason for removing CLSCompliant attributes?

niemyjski:
Thanks for the updated powershell script! I agree with Eric on renaming the ignore attribute to `ExceptionlessIgnoreAttribute` Then you know you are specifically ignoring it only in exceptionless.

src/Exceptionless/update-json.ps1
yang-xiaodong: Because it not exposes externally visible types, so add this attribute will cause more warnings

yang-xiaodong:
I updateed , please check.

By the way , the code in [9.0.1.zip](https://github.com/JamesNK/Newtonsoft.Json/archive/9.0.1.zip) is inconsistent with our current source code (Newtonsoft.Json), I have not replaced it.

niemyjski:
Looks good thanks for the pr! I'm going to try updating us to the latest json.net and push a release soon. It would be nice to get some more changes in the next release.

PR #200, opened 2018-11-20
Update RequestInfoCollector.cs

lvse:
fix bug,When Post data is less than 1024, it does not display problems.


niemyjski:
Thanks for the PR and great find!

PR #205, opened 2019-02-04
Extend Exceptionless.AspNetCore

moogle001:
Changed ExceptionlessAspNetCorePlugin to allow for retrieving HttpContext via IHttpContextAccessor.

Made ExceptionlessAspNetCorePlugin, ExceptionlessDiagnosticListener, and RequestInfoCollector public classes.


ejsmith:
@moogle001 not sure if you saw the previous comment. Can you test to see what happens when the `IHttpContextAccessor` hasn't been registered? I'm guessing this class will blow up when we try to create it because the DI container won't be able to find the dependency. We can probably inject `IServiceProvider` instead and try to resolve it dynamically.

niemyjski:
@moogle001 were you able to see if we can use the IServiceProvider to gracefully handle when IHttpContextAccessor is not registered?

niemyjski:
Thanks for the PR

moogle001:
Sorry, missed the email notifications of your response.

`var contextAccessor = app.ApplicationServices.GetService<IHttpContextAccessor>();`

This will return null if IHttpContextAccessor is not registered. The new constructor for ExceptionlessAspNetCorePlugin allows it to be null, so it will not produce errors, but will not be able to retrieve the session info.

PR #220, opened 2020-01-29
Windows & WPF .NET Core 3.1 support & deprecate signed packages

niemyjski:
@rwecho * add winform & wpf for dotnet core 3.0 preview.

@rwecho * add nuget script

@rwecho * remove core projects for dotnet core 3.0 and will replaced with fx version.

@rwecho * replace fx version with multi targeting

@rwecho * fix obsolete namepsace.

@rwecho * upgrade .NET Core 3.1

@rwecho* restore frameworks portable46-net451+win81+wpa81


niemyjski:
@rwecho is there any chance you could take a look at the package step of the build It seems broken. Otherwise I'll try and take a look over the next several days.

niemyjski:
@rwecho or could you take a look at why the sample winforms app is crashing when running on .net 4.5

PR #227, opened 2020-03-17
Removed Portable Package (PCL) support

niemyjski:
We may also want to consider bumping the required .NET Framework version to the latest supported as well or at least a newer version so we can start using newer compiler features without worrying about it. https://dotnet.microsoft.com/platform/support/policy/dotnet-framework


ejsmith:
Lots of deleted code makes me happy. 😀

Issue #233, opened 2020-07-24
Got wrong MachineName when using Exceptionless client v4.4.0


witskeeper:
Got wrong MachineName when using Exceptionless client v4.4.0 with ASP.NET Core 3.1.x

Exceptionless  v4.3.x works well.

niemyjski:
What is the difference between process machine name and the environment machine name for you? Also, I'm wondering if we should be filling back to look it up from the process if the environment machine name is empty. Looks like Process.MachineName will return the computer name where the process is running on . https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.machinename?view=netcore-3.1 Makes me think we should be using the process instead of the actual machine name.

witskeeper:
this bug came from v4.4.0  and you can see the  changes of code.
old version use `process.MachineName` when `#elif !PORTABLE && !NETSTANDARD1_2`

So , we can just use `Environment.MachineName` now. 

![DE422C88-80A7-4173-94B9-169D7000CCAD](https://user-images.githubusercontent.com/5594094/88610784-1dc27880-d0ba-11ea-9b04-301d0f051583.png)

niemyjski:
Thanks for the PR!

witskeeper:
when will the new version be released ？

niemyjski:
We like to try and get a few changes out in each release, can you use a nightly build for now?

yuefengkai:
If this issues is resolved, please notify me, thank you @niemyjski

niemyjski:
this issue is resolved and merged in, you can find the nuget packages on our nightly ci feed: https://f.feedz.io/exceptionless/exceptionless/nuget/index.json

niemyjski:
I'll also release a new version sometime today.

niemyjski:
A new release is building right now: https://github.com/exceptionless/Exceptionless.Net/releases/tag/v4.4.1

PR #234, opened 2020-07-24
fix:wrong  MachineName in v4.4.0

witskeeper:
fix:wrong  MachineName in v4.4.0 

see issues #233

### DIFF
--- a/src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
+++ b/src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
@@ -145,8 +145,6 @@ namespace Exceptionless.Services {
             if (_config.IncludeMachineName) {
                 try {
                     info.MachineName = Environment.MachineName;
-                    Process process = Process.GetCurrentProcess();
-                    info.MachineName = process.MachineName;
                 } catch (Exception ex) {
                     _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get machine name. Error message: {0}", ex.Message);
                 }


src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
```diff
@@ -145,8 +145,6 @@ public class DefaultEnvironmentInfoCollector : IEnvironmentInfoCollector {
             if (_config.IncludeMachineName) {
                 try {
                     info.MachineName = Environment.MachineName;
-                    Process process = Process.GetCurrentProcess();
```
niemyjski, line 148: Nice catch on it doing both of these (it's kind of weird). I'm thinking we would want to use the Process.MachineName (https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.machinename?view=netcore-3.1) which returns the machine name that the process is running on (in case of executing processes on remote machines) and fallback to Environment Machine name if some case it's null. Thoughts @witskeeper @ejsmith

ejsmith: I have no idea why we would want to use machine name from process. I've never seen that used and seems very strange to me. Why was it added?

witskeeper: Environment.MachineName return pod name in docker.
it's usefull  in exceptionless dashboard

witskeeper: In this case，we should use `Environment.MachineName` case  process always  executing on the local machine.
`Process.MachineName` return   “.”  and it's useless.

niemyjski:
Thanks for the PR!

PR #236, opened 2020-09-29
Improve Async stack traces, Improve Version Handling, Don't get Post Data on GETs, Update web pack

### LINKED ISSUE #177: Use `Ben.Demystifier` to produce better stack traces
I've noticed when I send in a "demystified" stack trace it doesn't quite display as cleanly as I'd like in the exceptionless web UI.  Maybe Exceptionless.Net could use this library internally (if the exception isn't already demystified)

https://github.com/benaadams/Ben.Demystifier

niemyjski:
This fixes #177


src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
```diff
@@ -45,7 +45,7 @@ public static class RequestInfoCollector {
             if (config.IncludeQueryString)
                 info.QueryString = context.Request.Query.ToDictionary(exclusionList);
 
-            if (config.IncludePostData)
+            if (!context.Request.Method.Equals("GET", StringComparison.OrdinalIgnoreCase) && config.IncludePostData)
```
niemyjski, line 48: @ejsmith I think this needs to be reverted or moved down. GetPostData gets form content. We also didn't update the other implementations.

src/Exceptionless/Exceptionless.csproj
```diff
@@ -38,6 +38,11 @@
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
```
niemyjski, line 43: @ejsmith are these also full framework packages?

src/Platforms/Exceptionless.AspNetCore/RequestInfoCollector.cs
ejsmith: That’s the entire point. It’s a GET. It doesn’t have post data.

ejsmith: I would agree that would should flip those conditions and maybe add some null checks to this. Also, if there are other parts that are getting post data on a GET then we should remove it there as well.

PR #243, opened 2021-01-06
Adding better support for serverless functions and sample projects to show how to use them

ejsmith:


samples/Exceptionless.SampleHosting/Exceptionless.SampleHosting.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
niemyjski, line 4: Any reason for not just targeting net5.0 as that is what will be downloaded from ms website?

samples/Exceptionless.SampleHosting/appsettings.json
```diff
@@ -0,0 +1,24 @@
+{
+  "Exceptionless": {
+    "ApiKey": "LhhP1C9gijpSKCslHHCvwdSIz298twx271n1l6xw",
+    "ServerUrl": "http://localhost:50000",
```
niemyjski, line 4: We probably should change all the server url ports in all sample apps to `5000`

src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
```diff
@@ -403,6 +403,21 @@ public static class ExceptionlessConfigurationExtensions {
...
+            string processQueueOnCompletedRequestValue = GetEnvironmentalVariable("Exceptionless:ProcessQueueOnCompletedRequest") ??
+                GetEnvironmentalVariable("Exceptionless__ProcessQueueOnCompletedRequest");
+
+            // if we are running in a serverless environment default this config to true
+            if (String.IsNullOrEmpty(processQueueOnCompletedRequestValue)
+                && (
+                    !String.IsNullOrEmpty(GetEnvironmentalVariable("AWS_EXECUTION_ENV"))
+                    || !String.IsNullOrEmpty(GetEnvironmentalVariable("FUNCTIONS_WORKER_RUNTIME")))
+                )
+                processQueueOnCompletedRequestValue = "true";
```
niemyjski, line 416: Might be better to use the `Bool.TrueString` constant. If we refactored this into a method, we could short circuit with a return without parsing the string.

src/Platforms/Exceptionless.Extensions.Hosting/ExceptionlessLifetimeService.cs
```diff
@@ -8,6 +8,9 @@ public class ExceptionlessLifetimeService : IHostedService {
 
         public ExceptionlessLifetimeService(ExceptionlessClient client, IHostApplicationLifetime appLifetime) {
             _exceptionlessClient = client;
+            
+            _exceptionlessClient.RegisterAppDomainUnhandledExceptionHandler();
```
niemyjski, line 12: If startup is called on the client I think these two handlers are already configured.

src/Exceptionless/Configuration/ExceptionlessConfiguration.cs
```diff
@@ -267,6 +267,12 @@ public class ExceptionlessConfiguration {
         /// </summary>
         public bool SessionsEnabled { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to automatically process the pending event queue after each request is completed.
+        /// NOTE: Only supported on AspNetCore currently.
+        /// </summary>
+        public bool ProcessQueueOnCompletedRequest { get; set; }
```
niemyjski, line 274: I wish this had a better name being In core configuration and tied to a web request, but I'm at a loss what a butter one would be.

niemyjski (commented):
Some good changes, left some feedback.

src/Platforms/Exceptionless.Extensions.Hosting/ExceptionlessLifetimeService.cs
ejsmith: That's fine they can be called multiple times. Not all scenarios call Startup. In the hosting one we are only calling this.

samples/Exceptionless.SampleHosting/Exceptionless.SampleHosting.csproj
ejsmith: Because nothing else targets .net 5.0. There is no need or benefit to changing this to latest.

samples/Exceptionless.SampleHosting/appsettings.json
ejsmith: Done

Issue #246, opened 2021-03-11
The FolderObjectStorage class SaveObject<T>method has a issues


d18zj:
When log level is changed from DEBUG to INFO, The client-side **server-settings.json** file will be incorrect.
``

{"@@UserAgentBotPatterns":"*bot*,*crawler*,*spider*,*aolbuild*,*teoma*,*yahoo*","@@log:*":"debug"}
``

After the update, there is an additional **}** sign at the end

`{"@@UserAgentBotPatterns":"*bot*,*crawler*,*spider*,*aolbuild*,*teoma*,*yahoo*","@@log:*":"info"}}`

should modify FolderObjectStorage.cs like this:
![image](https://user-images.githubusercontent.com/4657391/110778738-b4385880-829d-11eb-98b5-eba82dba1061.png)

niemyjski:
Thanks for this report and nice find! I wonder why we did OpenWrite there. Is there any chance you could create a test and submit this change in a pr?

PR #247, opened 2021-03-12
Fixed issues #246 FolderObjectStorage remaining the last content when the saved 

d18zj:
…ontent is reduced

### DIFF
--- a/src/Exceptionless/Storage/FolderObjectStorage.cs
+++ b/src/Exceptionless/Storage/FolderObjectStorage.cs
@@ -68,7 +68,7 @@ namespace Exceptionless.Storage {
                 if (!Directory.Exists(directory))
                     Directory.CreateDirectory(directory);
 
-                using (var writer = File.OpenWrite(Path.Combine(Folder, path))) {
+                using (var writer = File.Open(Path.Combine(Folder, path), FileMode.Create)) {
                     _resolver.GetStorageSerializer().Serialize(value, writer);
                 }
             } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


src/Exceptionless/Storage/FolderObjectStorage.cs
```diff
@@ -68,7 +68,7 @@ public class FolderObjectStorage : IObjectStorage {
                 if (!Directory.Exists(directory))
                     Directory.CreateDirectory(directory);
 
-                using (var writer = File.OpenWrite(Path.Combine(Folder, path))) {
+                using (var writer = File.Open(Path.Combine(Folder, path), FileMode.Create)) {
```
ejsmith, line 71: So previously it was appending content to the file if it already existed and this will always overwrite it?

test/Exceptionless.Tests/Storage/FolderFileStorageTests.cs
```diff
@@ -16,18 +18,49 @@ public class FolderFileStorageTests : FileStorageTestsBase {
...
+           
             var storage = new FolderObjectStorage(resolver, DATA_DIRECTORY_QUEUE_FOLDER);
             Assert.NotNull(storage.Folder);
             Assert.NotEqual(DATA_DIRECTORY_QUEUE_FOLDER, storage.Folder);
             Assert.True(storage.Folder.EndsWith("Queue" + Path.DirectorySeparatorChar) || storage.Folder.EndsWith("Queue" + Path.AltDirectorySeparatorChar), storage.Folder);
         }
+
+
+        [Fact]
+        public void Get_Save_Object_Multiple_Test() {
```
ejsmith, line 39: This test was failing before this change?

d18zj: Yes, GetObject will deserialize an exception and return null.

d18zj: Fixed issues #246

ejsmith:
Thank you!

PR #249, opened 2021-03-20
Upgraded packages to use .NET 5.0

niemyjski:
This is a breaking change for the WinForms and WPF nuget packages as it does drop .net core 3.1 in favor of .NET 5


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.100",
+    "version": "5.0.100",
     "rollForward": "latestMinor"
```
niemyjski, line 4: @ejsmith we were doing `latestMajor` in exceptionless.

benaadams:
Bump `net452` and `v4.5.2` to `net462` and `v4.6.2` also?

benaadams:
> Bump `net452` and `v4.5.2` to `net462` and `v4.6.2` also?

Made PR for that https://github.com/exceptionless/Exceptionless.Net/pull/250

global.json
ejsmith: Yeah, latestMinor is probably the correct choice. That's what we switched to in other projects.

niemyjski:
@benaadams Thanks!

PR #257, opened 2021-04-27
Lock conflicts occur when multiple sites (with ExceptionLess.net) are deployed on a server

d18zj:
Mutex is a cross-process lock,Lock conflicts occur when multiple sites (with ExceptionLess.net) are deployed on a server


src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -157,10 +156,9 @@ public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
                     }
                 });
             } catch (Exception ex) {
-                System.Diagnostics.Trace.WriteLine("Exceptionless: Error flushing log contents to disk: {0}", ex.Message);
+                System.Diagnostics.Trace.WriteLine("Exceptionless: Error flushing log contents to disk: {0}", ex.ToString());
```
niemyjski, line 159: Any specific reason for calling ToString() here instead of using the message?

niemyjski (approved):
This looks good to me, but good to get an extra pair of eyes on this! Thanks for the PR

src/Exceptionless/Logging/FileExceptionlessLog.cs
d18zj: The To string method returns more relevant information, such as the error stack.
In a real production environment, if more than one site is deployed on a single server, this lock conflict would cause memory depletion.I changed the code here to help me pinpoint the problem

ejsmith:
It's been a long time, but I'm pretty sure it's using a cross process lock on purpose. If you have multiple clients using the same log file then you need to make sure only one of them is appending to the file at the same time.

What behavior are you seeing that is causing issues? They should be retrying when they get conflicts.

niemyjski:
@d18zj could you please comment on @ejsmith's question.

niemyjski:
@d18zj could you please comment on @ejsmith's question.

niemyjski:
@d18zj could you please comment on @ejsmith's question.

ejsmith:
So I do see an issue with us taking a global cross process lock based on just the class name and not the log file we are trying to write to. We should change this to use a Mutex based on the file name like this implementation: https://github.com/bhaeussermann/NLog/blob/master/src/NLog/Internal/FileAppenders/MutexMultiProcessFileAppender.cs#L76

src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Exceptionless.Logging {
     public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
-        private static Mutex _flushLock = new Mutex(false, nameof(FileExceptionlessLog));
+        private static readonly object _flushLock = new object();
```
ejsmith, line 10: Change this to be based on the file name like this: https://github.com/bhaeussermann/NLog/blob/master/src/NLog/Internal/FileAppenders/MutexMultiProcessFileAppender.cs#L76

niemyjski:
Thanks for reporting this issue and creating a pr. I'm closing this in favor of https://github.com/exceptionless/Exceptionless.Net/pull/305

PR #263, opened 2021-11-30
Updated readme to point to exceptionless.com

deanhume:
Updated from exceptionless.io to exceptionless.com - it seems to hit the incorrect URL


niemyjski:
Thanks for the PR and bringing this to our attention. It looks like we have a dns issue which we are going to fix.  @ejsmith do you feel we should keep these on the .io domain or move them to .com?

niemyjski:
We've resolved the dns issue a few days ago but I think it's good idea to update this too.

PR #266, opened 2021-12-30
Internalize client extension methods

niemyjski:
This helps reduce external code dependencies


src/Exceptionless/Extensions/StringExtensions.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Exceptionless.Extensions {
     public static class StringExtensions {
```
niemyjski, line 7: @ejsmith We couldn't internalize the whole class because AnyWildCardMatches is used by at least 4 client implementations.

PR #270, opened 2022-04-23
Add ExceptionlessWindowsEnvironmentInfoCollector

### LINKED ISSUE #269: Add Exceptionless.Windows plugin to capture additional helpful information
Hi,

I wrote a plugin to capture the processes handle count, user object count, and GDI object count.

To reduce maintenance and improve portability. I think this information should be made available via a plugin in Exceptionless. Similar to the web/webapi plugins.

Would you accept a pull request for a plugin added to Exceptionless.Windows? 

This is what I currently use, but I imagine there is a nicer approach using dependency injection with IEnvironmentInfoCollector so that we can extend EnvironmentInfo? 

```vb
Public Class SysInfoPlugin
    Implements IEventPlugin
    Public Sub Run(context As EventPluginContext) Implements IEventPlugin.Run
        If context.Event.Type = [Event].KnownTypes.Error Then
            Dim info As Data.EnvironmentInfo = Nothing
            If context.Event.Data.TryGetValue([Event].KnownDataKeys.EnvironmentInfo, info) Then
                info.Data.Add("Handle Count", Process.GetCurrentProcess.HandleCount)
                info.Data.Add("User Object Count", GetGuiResourcesUserCount())
                info.Data.Add("GDI Object Count", GetGuiResourcesGDIObjectCount())
            End If
        End If
    End Sub

    <DllImport("User32")>
    Public Shared Function GetGuiResources(hProcess As IntPtr, uiFlags As Integer) As Integer
    End Function

    Public Shared Function GetGuiResourcesUserCount() As Integer
        Return GetGuiResources(Process.GetCurrentProcess().Handle, 1)
    End Function
    Public Shared Function GetGuiResourcesGDIObjectCount() As Integer
        Return GetGuiResources(Process.GetCurrentProcess().Handle, 0)
    End Function
End Class
```

Cheers,
Lachlan

elachlan:
Fixes #269

### DIFF
--- a/src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
+++ b/src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
@@ -96,21 +96,6 @@ namespace Exceptionless.Models.Data {
         public string RuntimeVersion { get; set; }
 
         /// <summary>
-        /// Gets the number of Handles used by the process.
-        /// </summary>
-        public int Handles { get; set; }
-
-        /// <summary>
-        /// Gets the number of User Objects used by the process.
-        /// </summary>
-        public int UserObjects { get; set; }
-
-        /// <summary>
-        /// Gets the number of GDI Objects used by the process.
-        /// </summary>
-        public int GDIObjects { get; set; }
-
-        /// <summary>
         /// Extended data entries for this machine environment.
         /// </summary>
         public DataDictionary Data { get; set; }
--- a/src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
+++ b/src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
@@ -96,6 +96,21 @@ namespace Exceptionless.Models.Data {
         public string RuntimeVersion { get; set; }
 
         /// <summary>
+        /// Gets the number of Handles used by the process.
+        /// </summary>
+        public int Handles { get; set; }
+
+        /// <summary>
+        /// Gets the number of User Objects used by the process.
+        /// </summary>
+        public int UserObjects { get; set; }
+
+        /// <summary>
+        /// Gets the number of GDI Objects used by the process.
+        /// </summary>
+        public int GDIObjects { get; set; }
+
+        /// <summary>
         /// Extended data entries for this machine environment.
         /// </summary>
         public DataDictionary Data { get; set; }
--- a/src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
+++ b/src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
@@ -13,15 +13,15 @@ using Exceptionless.Models.Data;
 namespace Exceptionless.Services {
     public class DefaultEnvironmentInfoCollector : IEnvironmentInfoCollector {
         private static EnvironmentInfo _environmentInfo;
-        private readonly ExceptionlessConfiguration _config;
-        private readonly IExceptionlessLog _log;
+        protected readonly ExceptionlessConfiguration config;
+        protected readonly IExceptionlessLog log;
 
         public DefaultEnvironmentInfoCollector(ExceptionlessConfiguration config, IExceptionlessLog log) {
-            _config = config;
-            _log = log;
+            this.config = config;
+            this.log = log;
         }
 
-        public EnvironmentInfo GetEnvironmentInfo() {
+        public virtual EnvironmentInfo GetEnvironmentInfo() {
             if (_environmentInfo != null) {
                 PopulateThreadInfo(_environmentInfo);
                 PopulateMemoryInfo(_environmentInfo);
@@ -38,43 +38,26 @@ namespace Exceptionless.Services {
             return _environmentInfo;
         }
 
-        private void PopulateApplicationInfo(EnvironmentInfo info) {
-            try {
-                info.Data.Add("AppDomainName", AppDomain.CurrentDomain.FriendlyName);
-            } catch (Exception ex) {
-                _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get AppDomain friendly name. Error message: {0}", ex.Message);
-            }
-
-            if (_config.IncludeIpAddress) {
-                try {
-                    IPHostEntry hostEntry = Dns.GetHostEntryAsync(Dns.GetHostName()).ConfigureAwait(false).GetAwaiter().GetResult();
-                    if (hostEntry != null && hostEntry.AddressList.Any())
-                        info.IpAddress = String.Join(", ", hostEntry.AddressList.Where(x => x.AddressFamily == AddressFamily.InterNetwork).Select(a => a.ToString()).ToArray());
-                } catch (Exception ex) {
-                    _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get ip address. Error message: {0}", ex.Message);
-                }
-            }
-        }
-
         private void PopulateProcessInfo(EnvironmentInfo info) {
             try {
                 info.ProcessorCount = Environment.ProcessorCount;
             } catch (Exception ex) {
-                _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get processor count. Error message: {0}", ex.Message);
+                log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get processor count. Error message: {0}", ex.Message);
             }
 
             try {
-                Process process = Process.GetCurrentProcess();
-                info.ProcessName = process.ProcessName;
-                info.ProcessId = process.Id.ToString(NumberFormatInfo.InvariantInfo);
+                using (Process process = Process.GetCurrentProcess()) {
+                    info.ProcessName = process.ProcessName;
+                    info.ProcessId = process.Id.ToString(NumberFormatInfo.InvariantInfo);
+                }
             } catch (Exception ex) {
-                _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get process name or id. Error message: {0}", ex.Message);
+                log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get process name or id. Error message: {0}", ex.Message);
             }
 
             try {
                 info.CommandLine = Environment.CommandLine;
             } catch (Exception ex) {
-                _log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get command line. Error message: {0}", ex.Message);
+                log.FormattedWarn(typeof(DefaultEnvironmentInfoCollector), "Unable to get command line. Error message: {0}", ex.Message);
             }
         }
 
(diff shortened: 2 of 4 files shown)


elachlan:
I think it might be a good idea to change the config/log fields to be protected so that inherited classes can log warnings and check the config without holding a reference themselves. This would change the API surface though, so I'd thought I'd raise it first.

elachlan:
If we lazy load `Process`, we need to implement a dispose method as we access `Process.Handle`.
https://stackoverflow.com/questions/16957320/what-does-process-dispose-actually-do

src/Platforms/Exceptionless.Windows/ExceptionlessWindowsEnvironmentInfoCollector.cs
```diff
@@ -0,0 +1,50 @@
...
+        private readonly IExceptionlessLog _log;
+        public ExceptionlessWindowsEnvironmentInfoCollector(ExceptionlessConfiguration config, IExceptionlessLog log) : base(config, log) {
+            _log = log;
+        }
+
+        public new EnvironmentInfo GetEnvironmentInfo() {
+            EnvironmentInfo info = base.GetEnvironmentInfo();
+            PopulateHandleInfo(info);
+            return info;
+        }
```
niemyjski, line 18: We should make this base class method virtual

niemyjski (changes requested):
Thanks for the PR, I will talk to the rest of the team about taking a dependency on win32. We do support this library on .NET 5+ even if it's windows only currently, it may not be some day. Also need to discuss if we feel like it's a good option to create new first-class properties that we would need to support everywhere. I feel like these properties probably should be in a data bag.

elachlan:
> Thanks for the PR, I will talk to the rest of the team about taking a dependency on win32. We do support this library on .NET 5+ even if it's windows only currently, it may not be some day. Also need to discuss if we feel like it's a good option to create new first-class properties that we would need to support everywhere. I feel like these properties probably should be in a data bag.

I'd like to hope that one-day winforms would run on platforms other than windows. I did ask Microsoft on GitHub about it. But was told winforms will forever be windows only due to its tight coupling with win32.

In WINE calls to these APIs will throw not implemented exceptions I think.

In any case, we can detect the platform and only populate the variables when it's running on windows. Handles are windows specific and so wouldn't make sense to attempt to be populated on Linux.

src/Exceptionless/Models/Client/Data/EnvironmentInfo.cs
```diff
@@ -95,6 +95,21 @@ public class EnvironmentInfo : IData {
...
+        /// <summary>
+        /// Gets the number of User Objects used by the process.
+        /// </summary>
+        public int UserObjects { get; set; }
+
+        /// <summary>
+        /// Gets the number of GDI Objects used by the process.
+        /// </summary>
+        public int GDIObjects { get; set; }
+
```
ejsmith, line 112: Definitely agree that this shouldn't be part of this class that is shared across a bunch of platforms. Moving them to data would work.

src/Platforms/Exceptionless.Windows/ExceptionlessWindowsEnvironmentInfoCollector.cs
ejsmith: Agreed

ejsmith (changes requested):
Thanks for the contribution @elachlan! Looks good, just a couple changes and we can get this merged.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
```diff
@@ -38,66 +38,50 @@ public class DefaultEnvironmentInfoCollector : IEnvironmentInfoCollector {
             return _environmentInfo;
         }
 
-        private void PopulateApplicationInfo(EnvironmentInfo info) {
```
elachlan, line 41: This was dead code.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
```diff
@@ -13,15 +13,15 @@
 namespace Exceptionless.Services {
     public class DefaultEnvironmentInfoCollector : IEnvironmentInfoCollector {
         private static EnvironmentInfo _environmentInfo;
-        private readonly ExceptionlessConfiguration _config;
-        private readonly IExceptionlessLog _log;
+        protected readonly ExceptionlessConfiguration config;
```
elachlan, line 16: It made sense to expose these to inherited classes as the constructor requires them to begin with.

elachlan:
> Thanks for the contribution @elachlan! Looks good, just a couple changes and we can get this merged.

Changes are applied, let me know what you both think.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
ejsmith: These should still be prefixed with an `_` as that is our naming convention.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
```diff
@@ -13,15 +13,15 @@
...
+        protected readonly IExceptionlessLog log;
 
         public DefaultEnvironmentInfoCollector(ExceptionlessConfiguration config, IExceptionlessLog log) {
-            _config = config;
```
ejsmith, line 20: These should be reverted to use `_`

ejsmith (changes requested):

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
elachlan: No problem. Visual Studio was complaining at me that `_` should only be used when something is private.

elachlan: https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs3008?f1url=%3FappId%3Droslyn%26k%3Dk(CS3008)

elachlan: I'll switch it to private and expose a property to access them.

ejsmith: I'm not saying they should be private. They can be protected.

elachlan: Renaming them to underscore causes CS3008 warning. so its best to have them private with underscore and expose them via a protected property.

Like this:
```
private readonly ExceptionlessConfiguration _config;
private readonly IExceptionlessLog _log;
protected ExceptionlessConfiguration Config => _config;
protected IExceptionlessLog Log => _log;
```

ejsmith: Hmm... I guess the `CLSCompliant` attribute is doing it. They do recommend turning them into protected properties. I guess just make them protected auto properties without a backing field.

elachlan: So its okay without the leading underscore?

ejsmith: No, was thinking like this:

```csharp
        public DefaultEnvironmentInfoCollector(ExceptionlessConfiguration config, IExceptionlessLog log) {
            Config = config;
            Log = log;
        }

        protected ExceptionlessConfiguration Config { get; }
        protected IExceptionlessLog Log { get; }
```

elachlan: How is that?

ejsmith (dismissed):
LGTM thank you @elachlan

elachlan:
The only nitpick I have is that GetGuiResources has the ability to query the peak count on GDI and User objects. Which might be helpful. But as per the documentation for it, it isn't supported on older windows versions:
`Windows Server 2008, Windows Vista, Windows Server 2003 and Windows XP:  This value is not supported until Windows 7 and Windows Server 2008 R2.`

That being said, I don't think that matters because of the requirement for .NET 4.6.2 by this package. Which is Win 7 Minimum.

src/Platforms/Exceptionless.Windows/ExceptionlessWindowsExtensions.cs
```diff
@@ -18,6 +19,7 @@ public static class ExceptionlessWindowsExtensions {
                 throw new ArgumentNullException(nameof(client));
 
             client.Configuration.AddPlugin<SetEnvironmentUserPlugin>();
+            client.Configuration.Resolver.Register<IEnvironmentInfoCollector, ExceptionlessWindowsEnvironmentInfoCollector>();
```
niemyjski, line 22: Can we confirm if that the WPF extensions method doesn't call this.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
niemyjski: Can we see when this was removed (and link to commit) and hook it back up. Looking at this, we should be capturing this data.

elachlan: It was added in Jun-2016: https://github.com/exceptionless/Exceptionless.Net/commit/3fa990d553b2e01621899a4c1b18833b807ead7b

For whatever reason, it was added and never used. Similar information is pulled in as apart of `PopulateProcessInfo()`.

niemyjski: That's weird, can we hook them back up.

src/Platforms/Exceptionless.Windows/ExceptionlessWindowsExtensions.cs
elachlan: Exceptionless.WPF and Exceptionless.Windows are separate projects and don't have any dependencies on each other as far as I can tell.

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
elachlan: `PopulateApplicationInfo()` uses `AppDomain.CurrentDomain.FriendlyName`. This is almost effectively the same as `process.ProcessName` populating the processname in `PopulateProcessInfo()`. 

Additionally, I think the IP Address is populated server side. Do you think that might clash when we add a local IP Address?

elachlan: Based on these, I am fairly certain they have the same information.
https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.friendlyname?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.processname?view=net-6.0

elachlan: @niemyjski do you think it is still needed based on that information?

src/Platforms/Exceptionless.Windows/ExceptionlessWindowsExtensions.cs
elachlan: @niemyjski are you satisfied based on the information provided above?

src/Exceptionless/Services/DefaultEnvironmentInfoCollector.cs
niemyjski: Let's revert this change and wire it up.

elachlan: Done!

niemyjski (approved):

niemyjski:
Thanks for the PR!

elachlan:
@niemyjski thank you for the review! Are you able to do a release sometime soon?

niemyjski:
We typically like to let a nightly sit for a little bit (week or two), we have a nightly NuGet feed you can use.

PR #272, opened 2022-05-04
Improve handling of Exception<T>

### LINKED ISSUE #83: Improve handling of Exception<T> where T : struct
@corneliutusnea commented on [Wed Aug 12 2015](https://github.com/exceptionless/Exceptionless/issues/132)

We use an Exception<T> which gets rendered a bit oddly in the UI.
public class CustomException<T> : ApplicationException
where T : struct
We use this so we can use enums instead of strings for error messages that are then easy to group in "buckets" or localize.

The the EL UI they get rendered as:
My.Custom.Namespace..CustomException`1[[Some.Long.Name.Space.ErrorCategory_ErrorBucket, Long.AssemblyName., Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]

EL should understand such exceptions and render them as:
CustomException<ErrorCategory_ErrorBucket>

---

@niemyjski commented on [Thu Aug 13 2015](https://github.com/exceptionless/Exceptionless/issues/132#issuecomment-130735046)

Yeah, I agree this should use the namespace+typename and not the assembly full name.

---

@niemyjski commented on [Thu Aug 13 2015](https://github.com/exceptionless/Exceptionless/issues/132#issuecomment-130737151)

It also looks like the grids are having issues with these type names too.

elachlan:
Fixes #83.

Reference: https://stackoverflow.com/questions/17480990/get-name-of-generic-class-without-tilde

### DIFF
--- a/src/Exceptionless/Extensions/ToErrorModelExtensions.cs
+++ b/src/Exceptionless/Extensions/ToErrorModelExtensions.cs
@@ -44,7 +44,7 @@ namespace Exceptionless {
 
             var error = new Error {
                 Message = exception.GetMessage(),
-                Type = type.FullName
+                Type = type.GetRealTypeName()
             };
 
             if (!isInner)
--- a/src/Exceptionless/Extensions/TypeExtensions.cs
+++ b/src/Exceptionless/Extensions/TypeExtensions.cs
@@ -118,5 +118,22 @@ namespace Exceptionless.Extensions {
                 || type == typeof(double)
                 || type == typeof(decimal);
         }
+
+        public static string GetRealTypeName(this Type t) {
+            if (!t.IsGenericType)
+                return t.FullName;
+
+            System.Text.StringBuilder sb = new System.Text.StringBuilder();
+            sb.Append(t.FullName.Substring(0, t.FullName.IndexOf('`')).Replace("+","."));
+            sb.Append('<');
+            bool appendComma = false;
+            foreach (Type arg in t.GetGenericArguments()) {
+                if (appendComma) sb.Append(',');
+                sb.Append(GetRealTypeName(arg));
+                appendComma = true;
+            }
+            sb.Append('>');
+            return sb.ToString();
+        }
     }
 }
--- a/src/Exceptionless/Extensions/ToSimpleErrorModelExtensions.cs
+++ b/src/Exceptionless/Extensions/ToSimpleErrorModelExtensions.cs
@@ -38,7 +38,7 @@ namespace Exceptionless.Extensions {
 
             var error = new SimpleError {
                 Message = GetMessage(exception),
-                Type = type.FullName,
+                Type = type.GetRealTypeName(),
                 StackTrace = exception.Demystify().StackTrace
             };
 
--- a/test/Exceptionless.Tests/Plugins/020_ErrorPluginTests.cs
+++ b/test/Exceptionless.Tests/Plugins/020_ErrorPluginTests.cs
@@ -260,5 +260,32 @@ namespace Exceptionless.Tests.Plugins {
                 Assert.Equal("{\"RandomValue\":\"Test\"}", json);
             }
         }
+
+        [Fact]
+        public void WillUnwrapExceptionTypeName() {
+            const string type = "Exceptionless.Tests.Plugins.PluginTestBase.GenericException<Exceptionless.Tests.Plugins.PluginTestBase.GenericException<System.String>>";
+
+            var errorPlugins = new List<IEventPlugin> {
+                new ErrorPlugin(),
+                new SimpleErrorPlugin()
+            };
+
+            var client = CreateClient();
+            foreach (var plugin in errorPlugins) {
+                var context = new EventPluginContext(client, new Event());
+                var exception = new GenericException<string>("Oops!");
+                var nestedException = new GenericException<GenericException<string>>(exception, "Oops, Again!");
+                context.ContextData.SetException(nestedException);
+                plugin.Run(context);
+                Assert.False(context.Cancel);
+                Assert.Equal(Event.KnownTypes.Error, context.Event.Type);
+                var error = context.Event.GetError();
+                if (error != null)
+                    Assert.Equal(type, error.Type);
+                else
+                    Assert.Equal(type, context.Event.GetSimpleError().Type);
+            }
+        }
+
     }
 }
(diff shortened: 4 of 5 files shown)


niemyjski (changes requested):
Thanks for the PR, can we also add test cases specifically addressed in the issue around structs. I also left some feedback about performance concerns

@ejsmith should we be worried about this creating lots of new stacks due to hash changes for type names?

elachlan:
How about this?
```C#
private static readonly ConcurrentDictionary<Type, string> TypeNameCache = new ConcurrentDictionary<Type, string>();
public static string GetRealTypeName(this Type t) {
    if (TypeNameCache.TryGetValue(t, out string name)) {
        return name;
    }
    if (!t.IsGenericType)
        return t.FullName.Replace('+','.');

    StringBuilder sb = new StringBuilder();
    ReadOnlySpan<char> fullName = t.FullName.AsSpan();
    int plusIndex = fullName.IndexOf('+');
    if (plusIndex > 0) {
        sb.Append(fullName.Slice(0, plusIndex).ToArray());
        sb.Append('.');
    }
    int length = fullName.IndexOf('`') - (plusIndex > 0 ? plusIndex + 1 : 0);
    sb.Append(fullName.Slice(plusIndex > 0 ? plusIndex + 1 : 0, length).ToArray());
    sb.Append('<');
    bool appendComma = false;
    foreach (Type arg in t.GetGenericArguments()) {
        if (appendComma) { sb.Append(','); }
        sb.Append(GetRealTypeName(arg));
        appendComma = true;
    }
    sb.Append('>');
    return TypeNameCache.GetOrAdd(t, sb.ToString());
}
```

elachlan:
Should I just use the TypeNameHelper?https://github.com/exceptionless/Exceptionless.Net/blob/118b006c1c66c7fcc2c668f91af5a27cad8bd2ab/src/Exceptionless/Demystifier/TypeNameHelper.cs#L48-L53

Only issue I see is that it includes the `+`, so instead of:
`Exceptionless.Tests.Plugins.PluginTestBase.GenericException<Exceptionless.Tests.Plugins.PluginTestBase.ErrorCategory>`
You get:
`Exceptionless.Tests.Plugins.PluginTestBase+GenericException<Exceptionless.Tests.Plugins.PluginTestBase+ErrorCategory>`

If this is okay, I will use that. I also suggest we add a cache of some kind to `TypeNameHelper`.

elachlan:
@niemyjski How is that? Do you want a couple more test cases?

src/Exceptionless/Demystifier/TypeNameHelper.cs
```diff
@@ -214,5 +231,20 @@ public DisplayNameOptions(bool fullName, bool includeGenericParameterNames)
 
             public bool IncludeGenericParameterNames { get; }
         }
+
+        private readonly struct DisplayNameCacheKey : IEquatable<DisplayNameCacheKey> {
+            public DisplayNameCacheKey(Type type, DisplayNameOptions displayNameOptions) {
+                TypeFullName=type.FullName;
+                DisplayNameOptions=displayNameOptions;
+            }
+            public string TypeFullName { get; }
```
elachlan, line 240: Maybe this can be the `Type` and we can use `Type.GetHashCode()` in the equals?

src/Exceptionless/Demystifier/TypeNameHelper.cs
```diff
@@ -38,6 +39,15 @@ public static class TypeNameHelper
...
         };
 
+        private static readonly ConcurrentDictionary<DisplayNameCacheKey, string> TypeFullDisplayNameCache = new();
+
+        /// <summary>
+        /// Gets the types pretty print full name.
+        /// </summary>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        public static string GetTypeFullDisplayName(this Type type) => GetTypeDisplayName(type, true, true);
```
elachlan, line 49: This is probably redundant. But I liked the Extension method.

elachlan:
@ejsmith any chance of a quick review?

src/Exceptionless/Demystifier/TypeNameHelper.cs
niemyjski, line 49: I don't think we should be modifying Demystifier at all, this is a third-party package we are source embedding. I also don't think we need to be caching this, as this adds a lot of complexity like cache eviction, memory management, possibly dead locking, etc.

If anything, these changes should be proposed upstream.

src/Exceptionless/Extensions/ToErrorModelExtensions.cs
```diff
@@ -41,10 +41,11 @@ internal static class ToErrorModelExtensions {
         private static Error ToErrorModelInternal(Exception exception, ExceptionlessClient client, bool isInner = false) {
             var log = client.Configuration.Resolver.GetLog();
             Type type = exception.GetType();
+            string typeName = type.GetTypeFullDisplayName();
```
niemyjski, line 44: Kinda like this being called `GetTypeDisplayName`

src/Exceptionless/Extensions/TypeExtensions.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text;
```
niemyjski, line 6: We can revert these unneeded changes

niemyjski (changes requested):
Left some comments, It's nice that we found an existing method that handles this!

elachlan:
> Left some comments, It's nice that we found an existing method that handles this!

All sorted. Thanks for the review!

niemyjski (approved):
Looks good

niemyjski:
Thanks for the PR!

niemyjski:
Thanks for the PR! These are really nice changes @elachlan was there any other changes you wanted to get in prior to a release?

elachlan:
> Thanks for the PR! These are really nice changes @elachlan was there any other changes you wanted to get in prior to a release?

I won't have the time to work on this for the next two weeks.

I was going to look at using spans in place of string operations at some point.

elachlan:
Is this the upstream?
https://github.com/benaadams/Ben.Demystifier/

niemyjski:
Yes

PR #274, opened 2022-05-09
Fix Exception.HResult is not being mapped to error.code property

### LINKED ISSUE #273: Exception.HResult is not being mapped to error.code property
`HResult` doesn't seem to be picked up properly via reflection in the ToError extensions. This could be due to it being a public property and the reflection is checking that it's private: https://docs.microsoft.com/en-us/dotnet/api/system.exception.hresult?msclkid=5283e541cf9b11ec938e45f7252349bc&view=net-6.0

> Starting with the .NET Framework 4.5, the [HResult](https://docs.microsoft.com/en-us/dotnet/api/system.exception.hresult?view=net-6.0) property's setter is protected, whereas its getter is public. In previous versions of the .NET Framework, both getter and setter are protected.

I'm thinking that since the library now targets .NET 4.5 and above, we can change our implementation to just check public.

```cs
    public class MyException : Exception {
        public TestOutOfMemoryException(string message) : base(message) {
            HResult = unchecked((int)0x8007000E);
        }
    }

    [Fact]
    private static void TrackException() {
        throw new TestOutOfMemoryException("Test");
        throw new OutOfMemoryException("Test");
    }

```

elachlan:
Fixes #273

### DIFF
--- a/src/Exceptionless/Extensions/ToErrorModelExtensions.cs
+++ b/src/Exceptionless/Extensions/ToErrorModelExtensions.cs
@@ -54,7 +54,7 @@ namespace Exceptionless {
             error.PopulateStackTrace(error, exception, log);
 
             try {
-                PropertyInfo info = type.GetProperty("HResult", BindingFlags.NonPublic | BindingFlags.Instance);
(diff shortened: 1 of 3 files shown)


niemyjski:
Awesome! Thank you for the PR!

PR #282, opened 2022-06-08
Make all Demystifier classes internal

### LINKED ISSUE #281: Updating from v4.6.2 to v4.7.0 or newer causes conflicts with Ben.Demystifier
In our project we were using Exceptionless v4.6.2. I just tried updating to the latest nuget package, and started getting errors becsuse the ToStringDemystified function was inaccessible.

'Exception' does not contain a definition for 'ToStringDemystified' and no accessible extension method 'ToStringDemystified' accepting a first argument of type 'Exception' could be found (are you missing a using directive or an assembly reference?)

We use ToStringDemystified in various places throughout our code for logging errors. We also use EnhancedStackTrace in a few places.

The problem is that now, as of v4.7.0, ExceptionExtensions, which contains ToStringDemystified, is declared internal in Exceptionless (see https://github.com/exceptionless/Exceptionless.Net/pull/266/files), but EnhancedStackTrace is still public.

So, if I use Exceptionless v4.7.0 or newer, I can no longer access ToStringDemystified, but if I add the Ben.Demystifier nuget package, then I get error code CS0433 stating:

The type 'EnhancedStackTrace' exists in both 'Ben.Demystifier, Version=0.4.0.0, Culture=neutral, PublicKeyToken=a6d206e05440431a' and 'Exceptionless, Version=4.0.0.0, Culture=neutral, PublicKeyToken=fc181f0a46f65747'

mlaboss-rsb:
Resolves #281 by making all of the classes from Ben.Demystifier internal instead of public.

### DIFF
--- a/src/Exceptionless/Demystifier/EnhancedStackFrame.cs
+++ b/src/Exceptionless/Demystifier/EnhancedStackFrame.cs
@@ -5,7 +5,7 @@ using System.Reflection;
 
 namespace System.Diagnostics
 {
-    public class EnhancedStackFrame : StackFrame
+    internal class EnhancedStackFrame : StackFrame
     {
         private readonly string? _fileName;
         private readonly int _lineNumber;
--- a/src/Exceptionless/Demystifier/EnhancedStackTrace.Frames.cs
+++ b/src/Exceptionless/Demystifier/EnhancedStackTrace.Frames.cs
@@ -17,7 +17,7 @@ using System.Threading.Tasks;
 
 namespace System.Diagnostics
 {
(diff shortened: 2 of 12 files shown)


ejsmith:
@mlaboss-rsb are there any newer updates to this library that we should include with this change?

mlaboss-rsb:
> @mlaboss-rsb thank you!! Are there any newer updates to this library that we should include with this change?

I don't believe so, the author of Demystifier incorporated updates in #248 and there haven't been any significant changes since then.

ejsmith (approved):

PR #283, opened 2022-07-27
Update DefaultSubmissionClient.cs

shamork:
WebRequestHandler: .net >=4.5
HttpClientHandler: .net >=4.5 OR netcore >=1.0 OR .NET Standard>=1.1
HttpClientHandler has better performance. So, we should use HttpClientHandler everywhere.


shamork:
WebRequestHandler open and close a tcp port every request and leaving it in TIME_WAIT state.
HttpClientHandler's keepalive feature reuse existing connections instead.

PR #284, opened 2022-07-27
Change System.Net.Http Framework Reference to PackageReference.

shamork:
Remove WebRequestHandler.
HttpClientHandler can do all the work with help of new System.Net.Http.

### DIFF
--- a/src/Exceptionless/Exceptionless.csproj
+++ b/src/Exceptionless/Exceptionless.csproj
@@ -39,6 +39,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <PackageReference Include="System.Reflection.Metadata" Version="6.0.1" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
   </ItemGroup>
@@ -51,8 +52,6 @@
     <Reference Include="Microsoft.VisualBasic" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Data" />
-    <Reference Include="System.Net.Http" />
-    <Reference Include="System.Net.Http.WebRequest" />
     <Reference Include="System.Numerics" />
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Xml" />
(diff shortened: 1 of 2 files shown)


src/Exceptionless/Exceptionless.csproj
```diff
@@ -39,6 +39,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
```
niemyjski, line 42: We don't want to take a package reference out on full framework when the assembly ships with .NET framework. As you can see item groups are conditional with references. Are you running into an issue where the type isn't in full framework?

shamork: System.Net.Http package supports all the dotnet runtimes and versions, and can be upgrade from nuget.
System.Net.Http from ships with .NET framework lacks ```HttpClientHandler.ServerCertificateCustomValidationCallback``` support for version <4.7.1.

niemyjski: Yes, but we don't want to take out a dependency that then integrators have to ship with their installers. We are targeting 4.6.2 currently (https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework) @ejsmith thoughts?

niemyjski: Yes, but we don't want to take out a dependency that then integrators have to ship with their installers. We are targeting 4.6.2 currently (https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework) 

If anything, I could see us adding additional targets for 471 and then not including this reference, but it really seems like we have no numbers or reproduction steps to fix this issue. I don't like fixing things that I can't reproduce or have documented proof they fix.

@ejsmith thoughts?

ejsmith: Yeah, I don't want to take dependencies if we don't have to. You can easily get into diamond dependency hell with such a commonly used package. We did that with JSON.NET and it was a nightmare.

I'm not really sure what the benefit is here considering if you are using a newer runtime, you will be using the package version of `System.Net.Http`. What benefit are you looking to get from thgis @shamork? If we provide a hook for customizing the HttpClient instance would that give you what you want?

niemyjski: @ejsmith can already customize it by replacing the `DefaultSubmissionClient` on the requested platforms.

ejsmith: True. Hopefully we make it easy to derive from ours and change it.

niemyjski:
We really appreciate the pr but cannot accept it at this time for the reasons above (taking a dependency on full framework). We've just merged in blazor support which makes the http client async.

PR #291, opened 2023-01-25
Add Windows environment info collector to the WPF platform

### LINKED ISSUE #290: Exceptionless.Windows vs Exceptionless.Wpf
I am a user of the Exceptionless.Wpf package. While looking at the source code for Exceptionless I stumbled upon [ExceptionlessWindowsEnvironmentInfoCollector.cs](https://github.com/exceptionless/Exceptionless.Net/blob/645a596f7b22baac2d4d22094e90ee247f625f89/src/Platforms/Exceptionless.Windows/ExceptionlessWindowsEnvironmentInfoCollector.cs). My app runs on Windows but I'm not seeing those process related properties in my exception reports. After some further investigation I understood that the Exceptionless.Windows package seems to be aimed at WinForms applications. The extension method [ExceptionlessWindowsExtensions.Register](https://github.com/exceptionless/Exceptionless.Net/blob/645a596f7b22baac2d4d22094e90ee247f625f89/src/Platforms/Exceptionless.Windows/ExceptionlessWindowsExtensions.cs#L17) wires up both WinForms exception handler and registers the ExceptionlessWindowsEnvironmentInfoCollector.

But I want those process handles, so I added the Exceptionless.Windows package but now I have a clash between the two Register extension methods. That can be solved, and I can register the collector with the resolver by hand. 

So, all in all I have probably worked around the problem but it really seems like the Wpf package should have the Windows environment collector by default. 

Maybe the packages should be split into:

- Exceptionless.Windows
- Exceptionless.WinForms or Exceptionless.Windows.WinForms
- Exceptionless.Wpf or Exceptionless.Windows.Wpf

mgnslndh:
Fixes #290 by add source link to `ExceptionslessWindowsEnvironmentInfoCollector` in the WPF platform project and registering that services to the dependency resolver.

              You should only need to use the Wpf client, code is shared between the clients. I'm trying to think if there is any reason we shouldn't be capturing that environment info in wpf. I can't think of any reason not to.. Would you mind submitting a pr to add a source link to that file in the wpf project and then adding a line to `client.Configuration.Resolver.Register<IEnvironmentInfoCollector, ExceptionlessWindowsEnvironmentInfoCollector>();` in register.

_Originally posted by @niemyjski in https://github.com/exceptionless/Exceptionless.Net/issues/290#issuecomment-1398800297_

### DIFF
--- a/src/Platforms/Exceptionless.Wpf/Exceptionless.Wpf.csproj
+++ b/src/Platforms/Exceptionless.Wpf/Exceptionless.Wpf.csproj
@@ -44,4 +44,8 @@
     <Resource Include="Images\ErrorFeedback.png" />
     <Folder Include="Properties\" />
   </ItemGroup>
+
+  <ItemGroup Label="Source Links">
+    <Compile Include="..\Exceptionless.Windows\ExceptionlessWindowsEnvironmentInfoCollector.cs" Link="ExceptionlessWindowsEnvironmentInfoCollector.cs" />
+  </ItemGroup>
 </Project>
--- a/src/Platforms/Exceptionless.Wpf/ExceptionlessWpfExtensions.cs
+++ b/src/Platforms/Exceptionless.Wpf/ExceptionlessWpfExtensions.cs
@@ -6,12 +6,13 @@ using Exceptionless.Dependency;
 using Exceptionless.Dialogs;
 using Exceptionless.Logging;
 using Exceptionless.Plugins.Default;
+using Exceptionless.Services;
 using Exceptionless.Wpf.Extensions;
 
 namespace Exceptionless {
     public static class ExceptionlessWpfExtensions {
         /// <summary>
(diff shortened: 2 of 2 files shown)


mgnslndh:
Not sure why the tests fail because they green on my machine :)

mgnslndh:
I have read the CLA Document and I hereby sign the CLA.

niemyjski:
@mgnslndh can you sign the cla, looks like a flakey test we need to look into.

mgnslndh:
Cloudflare is reporting the clahub as down. Not sure if it is affected by the service outage at Microsoft.
https://www.clahub.com/agreements/exceptionless/Exceptionless

I will sign as soon as I can when the site is up and running.

mgnslndh:
@niemyjski I've signed the CLA. The link was wrong but I've submitted another PR fixing that.

niemyjski:
Thanks for the PR!

PR #299, opened 2023-04-15
ExceptionlessTarget - ApiKey and ServerUrl and Message and UserIdentity with Layout support

### LINKED ISSUE #296: NLog: How can I send user identity?
Aspnet Web user cannot configured by NLog.config and write to el?
like this:

`<field name="UserIdentity" layout="${user}"/>`

### LINKED ISSUE #297: Update NLog ExceptionlessTarget to use Layout for ApiKey and ServerUrl
Change these lines:

https://github.com/exceptionless/Exceptionless.Net/blob/17841c9b07858cb3aa5f5f4d5cf038afff5a7bf2/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs#L12-L13

To these lines:

```c#
 public Layout ApiKey { get; set; } 
 public Layout ServerUrl { get; set; } 
```

Then update `InitializeTarget()`

```c#
        protected override void InitializeTarget() {
            base.InitializeTarget();

            var apiKey = RenderLogEvent(ApiKey, LogEventInfo.CreateNullEvent());
            var serverUrl = RenderLogEvent(ServerUrl, LogEventInfo.CreateNullEvent());

            if (!String.IsNullOrEmpty(apiKey) || !String.IsNullOrEmpty(serverUrl))
                _client = new ExceptionlessClient(config => {
                    if (!String.IsNullOrEmpty(apiKey) && apiKey != "API_KEY_HERE")
                        config.ApiKey = apiKey;
                    if (!String.IsNullOrEmpty(serverUrl))
                        config.ServerUrl = serverUrl;
                    config.UseInMemoryStorage();
                });
        }
```
This will enable[ NLog Target Configuration](https://github.com/NLog/NLog/wiki/Layouts#target-configuration-with-layout) where settings can be retrieved from app.config or appsettings.json (making it easier to have environment-specific setups)

snakefoot:
Resolves #297 and resolves #296

```xml
    <target type="Exceptionless, Exceptionless.NLog" name="exceptionless"
            apiKey="API_KEY_HERE"
            userIdentity="${aspnet-user-identity}"
            userIdentityName="${aspnet-user-claim:ClaimTypes.Name}">
      <field name="host" layout="${hostname}" />
      <field name="process" layout="${processname}" />
      <field name="user" layout="${environment-user}" />
    </target>
```
[${aspnet-user-identity}](https://github.com/NLog/NLog/wiki/AspNetUserIdentity-layout-renderer) comes from [NLog.Web.AspNetCore](https://www.nuget.org/packages/NLog.Web.AspNetCore)-nuget-package . See also: https://nlog-project.org/config/?tab=layout-renderers

### DIFF
--- a/src/Platforms/Exceptionless.NLog/ExceptionlessClientExtensions.cs
+++ b/src/Platforms/Exceptionless.NLog/ExceptionlessClientExtensions.cs
@@ -74,7 +74,7 @@ namespace Exceptionless.NLog {
             CreateFromLogEvent(client, ev).Submit();
         }
 
-        private static readonly HashSet<string> _ignoredEventProperties = new HashSet<string> {
+        private static readonly HashSet<string> _ignoredEventProperties = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
             "Tags",
             "ContextData"
         };
--- a/src/Platforms/Exceptionless.NLog/ExceptionlessClientExtensions.cs
+++ b/src/Platforms/Exceptionless.NLog/ExceptionlessClientExtensions.cs
@@ -8,6 +8,10 @@ using NLog;
 namespace Exceptionless.NLog {
     public static class ExceptionlessClientExtensions {
         public static EventBuilder CreateFromLogEvent(this ExceptionlessClient client, LogEventInfo ev) {
+            return CreateFromLogEvent(client, ev, ev.FormattedMessage);
+        }
+
+        public static EventBuilder CreateFromLogEvent(this ExceptionlessClient client, LogEventInfo ev, string formattedMessage) {
             if (client == null)
                 throw new ArgumentNullException(nameof(client));
 
@@ -20,10 +24,10 @@ namespace Exceptionless.NLog {
             builder.Target.Date = ev.TimeStamp;
             builder.SetSource(ev.LoggerName);
 
-            if (ev.Properties.Count > 0) {
+            if (ev.HasProperties) {
                 foreach (var property in ev.Properties) {
                     string propertyKey = property.Key.ToString();
-                    if (_ignoredEventProperties.Contains(propertyKey, StringComparer.OrdinalIgnoreCase))
+                    if (_ignoredEventProperties.Contains(propertyKey))
                         continue;
 
                     if (propertyKey.Equals("@value", StringComparison.OrdinalIgnoreCase)) {
@@ -53,8 +57,8 @@ namespace Exceptionless.NLog {
                 builder.SetType(Event.KnownTypes.Error);
             }
 
-            if (!String.IsNullOrWhiteSpace(ev.FormattedMessage))
-                builder.SetMessage(ev.FormattedMessage);
+            if (!String.IsNullOrWhiteSpace(formattedMessage))
+                builder.SetMessage(formattedMessage);
 
             var tags = ev.GetTags();
             if (tags != null)
@@ -70,10 +74,7 @@ namespace Exceptionless.NLog {
             CreateFromLogEvent(client, ev).Submit();
         }
 
-        private static readonly List<string> _ignoredEventProperties = new List<string> {
-            "CallerFilePath",
-            "CallerMemberName",
-            "CallerLineNumber",
+        private static readonly HashSet<string> _ignoredEventProperties = new HashSet<string> {
             "Tags",
             "ContextData"
         };
--- a/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
+++ b/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
@@ -13,6 +13,8 @@ namespace Exceptionless.NLog {
         public Layout ApiKey { get; set; }
         public Layout ServerUrl { get; set; }
         public Layout UserIdentity { get; set; }
+        public Layout UserIdentityName { get; set; }
+        public Layout UserIdentityEmail { get; set; }
 
         [ArrayParameter(typeof(ExceptionlessField), "field")]
         public IList<ExceptionlessField> Fields { get; set; }
@@ -50,8 +52,12 @@ namespace Exceptionless.NLog {
             var builder = _client.CreateFromLogEvent(logEvent, formattedMessage);
 
             var userIdentity = RenderLogEvent(UserIdentity, logEvent);
-            if (!String.IsNullOrWhiteSpace(userIdentity))
-                builder.Target.SetUserIdentity(userIdentity);
+            var userIdentityName = RenderLogEvent(UserIdentityName, logEvent);
+            builder.Target.SetUserIdentity(userIdentity, userIdentityName);
+
+            var userIdentityEmail = RenderLogEvent(UserIdentityEmail, logEvent);
+            if (!String.IsNullOrWhiteSpace(userIdentityEmail))
+                builder.SetUserDescription(userIdentityEmail, userIdentityName);
 
             foreach (var field in Fields) {
                 var renderedField = RenderLogEvent(field.Layout, logEvent);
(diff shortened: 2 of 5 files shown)


src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
```diff
@@ -40,9 +47,19 @@ public class ExceptionlessTarget : TargetWithLayout {
             if (logEvent.Level < minLogLevel)
                 return;
 
-            var builder = _client.CreateFromLogEvent(logEvent);
+            var formattedMessage = RenderLogEvent(Layout, logEvent);
+            var builder = _client.CreateFromLogEvent(logEvent, formattedMessage);
+
+            var referenceId = RenderLogEvent(ReferenceId, logEvent);
+            if (!String.IsNullOrWhiteSpace(referenceId))
+                builder.SetReferenceId(referenceId);
```
niemyjski, line 55: I don't think we want to set this like this. Reference id's are deduped server side so if you have multiple log messages in say a controller action, you are only going to get the first one.

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
```diff
@@ -40,9 +47,19 @@ public class ExceptionlessTarget : TargetWithLayout {
...
+
+            var userIdentity = RenderLogEvent(UserIdentity, logEvent);
+            if (!String.IsNullOrWhiteSpace(userIdentity))
+                builder.Target.SetUserIdentity(userIdentity);
```
niemyjski, line 59: Q: Is RenderLogEvent expensive? Also is there a way to get the friendly user name?

niemyjski (commented):
I left a comment. I don't think we can use reference id's like this because they will be discarded events server side as each event needs to be unique, can you please confirm.

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
niemyjski: I agree

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
snakefoot: Thought ReferenceId was similar to CorrelationId, so one could group LogEvents to the same request-chain together. But I will remove "ReferenceId".

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
snakefoot: I thought that User-Identification was a big no-no in the cloud-world, and even if logging the UserIdentity then it must be obfuscated upfront. But you are very wellcome to continue improving the NLog Target and adding even more awesome futures. I just wanted to show how I would resolve #296

src/Platforms/Exceptionless.NLog/ExceptionlessClientExtensions.cs
```diff
@@ -20,10 +24,10 @@ public static class ExceptionlessClientExtensions {
             builder.Target.Date = ev.TimeStamp;
             builder.SetSource(ev.LoggerName);
 
-            if (ev.Properties.Count > 0) {
+            if (ev.HasProperties) {
                 foreach (var property in ev.Properties) {
                     string propertyKey = property.Key.ToString();
-                    if (_ignoredEventProperties.Contains(propertyKey, StringComparer.OrdinalIgnoreCase))
+                    if (_ignoredEventProperties.Contains(propertyKey))
```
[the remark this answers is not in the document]
snakefoot: Yes will add it to the` HashSet<string> _ignoredEventProperties`

niemyjski:
> I left a comment. I don't think we can use reference id's like this because they will be discarded events server side as each event needs to be unique, can you please confirm.

@ejsmith may want to consider bringing correlation id in as it may help track down issues in otel. I'm not sure how well our event references we store in extended data for this scenario would help.

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
```diff
@@ -40,9 +48,19 @@ public class ExceptionlessTarget : TargetWithLayout {
...
+            var formattedMessage = RenderLogEvent(Layout, logEvent);
+            var builder = _client.CreateFromLogEvent(logEvent, formattedMessage);
+
+            var userIdentity = RenderLogEvent(UserIdentity, logEvent);
+            var userIdentityName = RenderLogEvent(UserIdentityName, logEvent);
+            builder.Target.SetUserIdentity(userIdentity, userIdentityName);
+
+            var userIdentityEmail = RenderLogEvent(UserIdentityEmail, logEvent);
+            if (!String.IsNullOrWhiteSpace(userIdentityEmail))
+                builder.SetUserDescription(userIdentityEmail, userIdentityName);
```
niemyjski, line 60: I think this is worth checking out more, may want to tweak or remove this later, typically is only used for a user entered description of the error.

src/Platforms/Exceptionless.NLog/readme.txt
```diff
@@ -36,12 +36,11 @@ minimum log level that will be used until the client retrieves settings from the
 <?xml version="1.0" encoding="utf-8" ?>
 <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <targets async="true">
-    <target xsi:type="Exceptionless, Exceptionless.NLog" name="exceptionless" apiKey="API_KEY_HERE">
+  <targets>
+    <target type="Exceptionless, Exceptionless.NLog" name="exceptionless" apiKey="API_KEY_HERE">
```
niemyjski, line 40: I think we still want these changes:
1. I don't know why we wouldn't want to use async wrapper here as it alleviates pressure on the application thread?: https://github.com/NLog/NLog/wiki/AsyncWrapper-target
2. They are still defining the xsi type here and every other page I checked (am I missing something)?: https://github.com/NLog/NLog/wiki/Getting-started-with-ASP.NET-Core-6
```suggestion
  <targets async="true">
    <target xsi:type="Exceptionless, Exceptionless.NLog" name="exceptionless" apiKey="API_KEY_HERE">
```

niemyjski (commented):
Looks great, had a question about target changes

src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
snakefoot: Doesn't sound like an approval, so will remove `userIdentityEmail` again, then it can be added back when requested by someone.

src/Platforms/Exceptionless.NLog/readme.txt
snakefoot: The idea with `xsi:type` is to support intellisense using NLog.Schema, but `xsi:type` doesn't allow use of special characters like comma (Ex. `, Exceptionless.NLog`), and will instead generate schema-warnings.

The idea with `async="true"` is to offload writing to a background-thread with support for batching, by writing to an in-memory-queu. But ExceptionLess-client already implements its own in-memory-queue with background-writer, so there is little benefit from adding `async="true"` (actually it introduces a small overhead without any reward)

snakefoot: Now see that your "queue" actually can live on isolated-storage / folder-storage, so I have restored `<targets async="true">`

niemyjski: Yes, you can call client.ProcessQueueAsync()

niemyjski (approved):

niemyjski:
Thanks for the PR!!!!!!

src/Platforms/Exceptionless.NLog/readme.txt
niemyjski: No, it should not. If you see an issue let us know.

snakefoot: Created #301 to perform flush on shutdown to avoid loosing logevents.

PR #300, opened 2023-04-18
Fixes #127: Added ability to record http headers

### LINKED ISSUE #127: Add ability to collect custom HTTP Headers
Asp.net WebApi Project using oAuth2 , the token info put in header:
`--header 'Authorization: Bearer 7f6b258211f24470b1384cd0614923df'`

Current Exceptionless.WebAPI not record this  header info

niemyjski:
Before merging this we probably need to make these changes server side as well. 

I also want to add MAX_DATA_ITEM_LENGTH to the other two implementations.

### DIFF
--- a/src/Exceptionless/Configuration/ExceptionlessConfiguration.cs
+++ b/src/Exceptionless/Configuration/ExceptionlessConfiguration.cs
@@ -228,6 +228,7 @@ namespace Exceptionless {
                 IncludeUserName = value;
                 IncludeMachineName = value;
                 IncludeIpAddress = value;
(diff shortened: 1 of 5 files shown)


ejsmith:
This looks good. Not sure how the string[] array will get handled by Exceptionless though. Should we just have it be a comma delimited string of header values?

niemyjski:
We need to update the backend first, but I'd rather do an array here than a comma delimited list as it would be easier to render out and then you have no issues escaping existing values that might have a comma.

niemyjski (commented):
Self reviewed

Issue #260, opened 2021-08-02
MinimumLevel setting is ignored while submitting the logs to dashboard


jbulowski:
Hello,

first of all i would like to apologize that I'm posting this issue here,
but the serilog exceptionless extension project doesn't include the issues page.

I'm initializing logger with serilog like following

```csharp
Log.Logger = new LoggerConfiguration()
                .MinimumLevel.Verbose()
                .WriteTo.Exceptionless(exceptionlessKey, exceptionlessUrl)
                .WriteTo.File(@"c:\logs\temp_logs.txt")
                .CreateLogger();
```

and while all the logs appear in the log file, I don't get logs below the `Warning` level in my dashboard.

I am using a self hosted version of excetionless and found the line that couses this 

https://github.com/serilog/serilog-sinks-exceptionless/blob/master/src/Serilog.Sinks.Exceptionless/Sinks/Exceptionless/ExceptionlessClientExtensions.cs#L33

On this line the `Constants.SourceContextPropertyName` isn't getting found and because of it, it defaults to `Warning` and  bypasses the inital setting (in `Exceptionless.Models.SettingsDictionary.GetMinLogLevel()`).

How do I fix it?

Thanks in advance

niemyjski:
That's really weird that there is no issues showing there, I'm not sure if the serilog team turned that off or not.  I think you may be running into the clients default minimum log level until server settings are set: https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Exceptionless/Configuration/ExceptionlessConfiguration.cs#L209-L216 Does calling this resolve this issue? Also do you have a default log level set in project settings?

I'd also look into why your SourceContextPropertyName isn't being set. Are you only injecting ILogger or ILogger<T>? Or are you not passing a name to ILoggerFactory.CreateLogger("name")?

jbulowski:
Calling `SetDefaultMinLogLevel` doesn't change anything.
And I'm only injecting ILogger like it's shown in the code snippet.

There is no settings in project setting for exceptionless.
On the server side, the MinimumLevel is set to `Debug`.

niemyjski:
Can you try enabling client logging and seeing what could be happening: https://exceptionless.com/docs/clients/dotnet/troubleshooting/

jbulowski:
I've enabled client logging and it's not logging anything into the specified file.
What I've found is that with the version `3.1.3` of the `Serilog.Sinks.Exceptionless` package everything works fine.
[...]

PR #303, opened 2023-04-27
NLog ExceptionlessTarget SetDefaultMinLogLevel

snakefoot:
See also: https://github.com/exceptionless/Exceptionless.Net/issues/260

### DIFF
--- a/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
+++ b/src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
@@ -39,6 +39,7 @@ namespace Exceptionless.NLog {
                         config.ServerUrl = serverUrl;
                     config.UseLogger(new NLogInternalLoggger());
                     config.UseInMemoryStorage();
+                    config.SetDefaultMinLogLevel(Logging.LogLevel.Trace);   // Rely on NLog Logging Rules
                 });
             }
             else {


src/Platforms/Exceptionless.NLog/ExceptionlessTarget.cs
```diff
@@ -39,6 +39,7 @@ public class ExceptionlessTarget : TargetWithLayout {
                         config.ServerUrl = serverUrl;
                     config.UseLogger(new NLogInternalLoggger());
                     config.UseInMemoryStorage();
+                    config.SetDefaultMinLogLevel(Logging.LogLevel.Trace);   // Rely on NLog Logging Rules
```
niemyjski, line 42: This works, but I wonder if there is a way to see what the min log level is for a given target? @ejsmith I kinda feel like maybe if you specify a logging target like nlog which doesn't have a restrict on the target itself, than this is the correct approach.

snakefoot: NLog Logger-output will never reach a NLog Target without first specifying a NLog Logging-Rule that redirects output from matching NLog Logger-names / LogLevels. See also https://github.com/NLog/NLog/wiki/Configuration-file#rules

niemyjski (approved):
Thanks for the PR!

snakefoot:
Maybe the override-logic of the call to `SetDefaultMinLogLevel` on the default-client should be: "override if not configured already" ?

niemyjski:
Yeah we could add an extension or check for that, wouldn't be a bad idea.

niemyjski:
@snakefoot I just looked at the implementation and it already handles this :)

```cs
        public void SetDefaultMinLogLevel(LogLevel minLogLevel) {
            if (!Settings.ContainsKey("@@log:*"))
                Settings.Add("@@log:*", minLogLevel.ToString());
        }
```

PR #305, opened 2023-05-03
Lock conflicts occur when multiple sites (with ExceptionLess.net) are deployed on a server

niemyjski:
Addresses https://github.com/exceptionless/Exceptionless.Net/pull/257

### DIFF
--- a/src/Exceptionless/Logging/FileExceptionlessLog.cs
+++ b/src/Exceptionless/Logging/FileExceptionlessLog.cs
@@ -1,19 +1,24 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Security;
+#if NET45
+using System.Security.AccessControl;
+using System.Security.Principal;
+#endif
 using System.Text;
 using System.Threading;
+using Exceptionless.Extensions;
 using Exceptionless.Utility;
 
 namespace Exceptionless.Logging {
     public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
-        private static Mutex _flushLock = new Mutex(false, nameof(FileExceptionlessLog));
-
+        private Mutex _flushMutex;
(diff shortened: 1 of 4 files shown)


src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -263,9 +268,14 @@ public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
             }
 
             Flush();
+
+            if (_flushMutex != null) {
+                _flushMutex.Close();
+                _flushMutex = null;
+            }
```
niemyjski, line 275: I added this, I don't see how it can hurt.

src/Exceptionless/Logging/FileExceptionlessLog.cs
```diff
@@ -298,6 +308,49 @@ public class FileExceptionlessLog : IExceptionlessLog, IDisposable {
...
+            } catch (Exception ex) {
+                if (ex is SecurityException || ex is UnauthorizedAccessException || ex is NotSupportedException || ex is NotImplementedException) {
+                    System.Diagnostics.Trace.WriteLine("Exceptionless: Error creating global mutex falling back to previous implementation: {0}", ex.ToString());
+                    return new Mutex(false, nameof(FileExceptionlessLog));
+                }
+
+                System.Diagnostics.Trace.WriteLine("Exceptionless: Error creating global mutex: {0}", ex.ToString());
+                throw;
+            }
+#else
```
niemyjski, line 333: Global mutexes are not supported on .NET Core or mono. There probably is a case we don't fully cover here.

ejsmith: According to the docs they are when the mutex name starts with Global\ which you are doing.

Issue #319, opened 2025-03-27
Unable to fully configure shared ExceptionlessClient.Default Instance via Web.config


nmdias:
Hi,

Has anyone successfully configured the shared instance of ExceptionlessClient using Web.config?

Legend has it that this should work.

```c#
ExceptionlessClient.Default.Configuration.ReadFromConfigSection();
```

My goal is to avoid duplicating configuration settings both in Web.config (which works well for unexpected exceptions and they are properly published) and in AppSettings. Ideally, I wouldn’t need to do this manually for exceptions that I catch myself:

```c#
ExceptionlessClient.Default.Configuration.Enabled = ApplicationSettings.ExceptionLessEnable;
ExceptionlessClient.Default.Startup(ApplicationSettings.ExceptionLessAppKey);
```

This would eliminate the need to define configurations separately in AppSettings.

**Expected Behavior**
	•	ExceptionlessClient.Default should read its configuration from Web.config automatically.
	•	No need to set values manually in code if they are already defined in Web.config.

**Actual Behavior**
	•	The configuration does not seem to be applied from Web.config, requiring explicit setup in code.

I'm using version 6, in a .NET 4.8 app.

For now, I also found a workaround with using `ConfigurationManager.GetSection("exceptionless")`, but it's forcing me to use reflection to access the properties, as the returned `ExceptionlessSection` is an internal class. 

Thank you

niemyjski:
Thanks for reaching out! Would you mind submitting a snippet of your web.config minus any extra configuration or api keys (blanked out). Also, compare your sample to our .NET Samples in this repo and see what is different.

What package do you have installed? 

Calling startup() by itself should resolve all of this. There are other things that come into play here that you need to take into account as configuration is layered. As in we read from (in order)

1. attribute
2. config section
3. app settings
4. env varaibles

I'd make sure that 3 or 4 are not overriding your settings.

nmdias:
Hi @niemyjski 

Thank you for pointing me to the samples, I made some progress from that.

I have these installed:

```
Exceptionless 6.0.4
Exceptionless.Mvc 6.0.4
```

So, before I had this in my Web.config, working for unexpected exceptions, but not for auto setup:

```xml
<configuration>
    <exceptionless apiKey="somekey" enabled="true" tags="tag1, tag2" />
</configuration>
```

Now I changed to this, except Tags that I'm unable to get:

```xml
<configuration>
[...]

PR #320, opened 2025-03-28
Get tags from appSettings #319

nmdias:


src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
```diff
@@ -390,6 +390,11 @@ public static void ReadFromAppSettings(this ExceptionlessConfiguration config) {
             string serverUrl = ConfigurationManager.AppSettings["Exceptionless:ServerUrl"];
             if (!String.IsNullOrEmpty(serverUrl))
                 config.ServerUrl = serverUrl;
+
+            string tags = ConfigurationManager.AppSettings["Exceptionless:Tags"];
```
niemyjski, line 394: @ejsmith any prerference here on Tags vs DefaultTags (this populates `ExceptionlessConfiguration.DefaultTags`)? 

- IConfiguration implementation looks for a section called `DefaultTags`
- ExceptionlessSection ConfigurationSection (legacy .NET) calls it `Tags`

niemyjski (commented):
Thanks for thre PR, it looks good just need to get to concensus on the name of it!

My vote is DefaultTags as that aligns with IConfiguration but open to feedback cc @ejsmith .

src/Exceptionless/Extensions/ExceptionlessConfigurationExtensions.cs
nmdias: Yes, i noticed the distinction, but to keep consistent with the already existing <exceptionless tags=''> in legacy .NET naming, I went  with Tags.

ejsmith: Yeah, seems like calling it DefaultTags is better describing what it does. I'd say if this is a new thing we are adding that we call it DefaultTags.

niemyjski (approved):

nmdias:
Super! Thanks :)

niemyjski:
New release is going out https://github.com/exceptionless/Exceptionless.Net/releases/tag/v6.1.0

nmdias:
Thank you :)

